PR #9, opened 2020-08-25
feat(button): ✨  add react aria buttons

navin-moorthy:
Road to add all React Aria components for our use case


navin-moorthy:
Added `Aria-Components` like `AriaButton` `AriaLink` and we export Reakit components directly.

src/Link/AriaLink.ts
```diff
@@ -0,0 +1,31 @@
...
+import { useForkRef } from "reakit-utils";
+import { useLink, AriaLinkOptions } from "@react-aria/link";
+import { BoxHTMLProps, useBox } from "reakit";
+import { mergeProps } from "@react-aria/utils";
+import { INTERACTION_KEYS } from "../interactions/__keys";
+import { createComponent, createHook } from "reakit-system";
+
+export const useAriaLink = createHook<AriaLinkOptions, BoxHTMLProps>({
+  name: "useAriaLink",
+  keys: [...INTERACTION_KEYS],
```
anuraghazra, line 11: No need to spread the keys here i guess.

src/button/AriaButton.ts
```diff
@@ -0,0 +1,32 @@
...
+import { BoxHTMLProps, useBox } from "reakit";
+import { useButton } from "@react-aria/button";
+import { mergeProps } from "@react-aria/utils";
+import { AriaButtonProps } from "@react-types/button";
+import { INTERACTION_KEYS } from "../interactions/__keys";
+import { createComponent, createHook } from "reakit-system";
+
+export const useAriaButton = createHook<AriaButtonProps, BoxHTMLProps>({
+  name: "useAriaButton",
+  keys: [...INTERACTION_KEYS],
```
anuraghazra, line 12: Same here. no need to spread.

src/button/AriaToggleButton.ts
```diff
@@ -0,0 +1,40 @@
...
+  BoxHTMLProps
+>({
+  name: "useAriaToggleButton",
+  keys: [...INTERACTION_KEYS, "isSelected", "setSelected", "toggle"],
+  compose: [useBox],
+
+  useProps(options, { ref: htmlRef, ...htmlProps }) {
+    const { isSelected, setSelected, toggle, ...restOptions } = options;
+    const props = { ...restOptions, ...htmlProps } as AriaToggleButtonProps;
+    const state = { isSelected, setSelected, toggle } as ToggleState;
```
anuraghazra, line 24: Is there any particular reason why we are using `as` assertion instead of using type declarations directly?

```ts
    const props: AriaToggleButtonProps = { ...restOptions, ...htmlProps };
    const state: ToggleState = {
      isSelected,
      setSelected,
      toggle,
    };
```

I can also see that using `const props: AriaToggleButtonProps = { ...restOptions, ...htmlProps };` typescript is complaining about some mismatched types.


<details>
<summary>Ts error</summary>

```ts
const props: AriaToggleButtonProps
Type '{ defaultChecked?: boolean | undefined; defaultValue?: string | number | readonly string[] | undefined; suppressContentEditableWarning?: boolean | undefined; suppressHydrationWarning?: boolean | undefined; ... 263 more ...; excludeFromTabOrder?: boolean | undefined; }' is not assignable to type 'AriaToggleButtonProps'.
  Types of property 'onChange' are incompatible.
    Type '((event: FormEvent<any>) => void) | ((isSelected: boolean) => void) | undefined' is not assignable to type '((isSelected: boolean) => void) | undefined'.
      Type '(event: FormEvent<any>) => void' is not assignable to type '(isSelected: boolean) => void'.
        Types of parameters 'event' and 'isSelected' are incompatible.
          Type 'boolean' is not assignable to type 'FormEvent<any>'.ts(2322)
```

</details>

src/button/index.ts
```diff
@@ -0,0 +1,3 @@
+export { Button, useButton } from "reakit/Button";
```
anuraghazra, line 1: `AriaButton` uses @react-aria/button right? why we are exporting reakit's button in here?

src/interactions/utils.ts
```diff
@@ -0,0 +1,15 @@
+// Keyboards, Assistive Technologies, and element.click() all produce a "virtual"
+// click event. This is a method of inferring such clicks. Every browser except
+// IE 11 only sets a zero value of "detail" for click events that are "virtual".
+// However, IE 11 uses a zero value for all click events. For IE 11 we rely on
+// the quirk that it produces click events that are of type PointerEvent, and
+// where only the "virtual" click lacks a pointerType field.
+
+export function isVirtualClick(event: MouseEvent | PointerEvent): boolean {
```
anuraghazra, line 8: Let's change the file name to __utils.js then, for consistency.

src/interactions/stories/Interactions.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { Keyboard } from "../Keyboard";
+import { FocusWithin } from "../FocusWithin";
+import { FocusVisible } from "../FocusVisible";
+import { Hover } from "../Hover";
+
+export default {
+  title: "Component/Interactions",
+} as Meta;
+
+export const Default = () => {
```
anuraghazra, line 15: Should we also create a story for each component? (Focus, Hover, Focusable, Keyboard etc etc)

anuraghazra:
Hey @navin-moorthy i've requested some minor changes and asked a few questions to clear up my confusions :) address them whenever you have some free time to do so.

src/button/index.ts
navin-moorthy: I exported it thinking that we are building a separate library.
Instead of using "reakit/Button" when using our components to build our UI.
I thought we export all the reakit components and use only from this repo, this way user need to only install our library.
Just a reexport and importing it again.

src/interactions/utils.ts
navin-moorthy: Done !!

src/interactions/stories/Interactions.stories.tsx
navin-moorthy: May be later, now we are only using it for our purpose.

I thought we can make these stories proper when we are writing the proper docs.

src/button/index.ts
anuraghazra: Okay got it! fair enough.

src/button/AriaToggleButton.ts
navin-moorthy: That is because the `useToggleButton` only accepts props of type `AriaToggleButtonProps`.

But `htmlPorps` has more types that are not accepted by `AriaToggleButtonProps`, so I had to override them.

`const props: AriaToggleButtonProps = { ...restOptions, ...htmlProps };`

By using directly, it doesn't match with the spreaded props, so had to use `as` to override it.

src/interactions/stories/Interactions.stories.tsx
anuraghazra: Okay!

navin-moorthy:
Merged with master to resolve conflicts and updated some packages to latest version.

Do check these and approve the PR to merge.

src/button/AriaToggleButton.ts
anuraghazra: Hmmm i see okay, i guess it's fine to use as assertion for this case.

anuraghazra:
LGTM!

anuraghazra (approved):

PR #10, opened 2020-08-26
wip: Experimental combobox

anuraghazra:


src/select/SelectDropdown.ts
```diff
@@ -28,16 +34,23 @@ const useSelectDropdown = createHook<SelectDropdownOptions, BoxHTMLProps>({
...
+      move,
+      setSelected,
+      setCurrentId,
+      setTypehead,
     },
     { ref: htmlRef, style: htmlStyle, ...htmlProps },
   ) {
+    const { handleOnKeyPress } = useTypeahead({
+      setTypehead,
+      disabled: isCombobox,
```
anuraghazra, line 51: Disabling typeahead in combobox while the dropdown is open, so that it does not interfare with combobox's input value.

src/select/SelectDropdown.ts
```diff
@@ -57,35 +70,36 @@ const useSelectDropdown = createHook<SelectDropdownOptions, BoxHTMLProps>({
...
+              }
             }
-          }
-        });
-      }, 150),
-      [typehead, visible],
-    );
+          });
+        },
+        visible ? 150 : 400,
```
anuraghazra, line 98: I'm still seeing the bug with multiselect so i'm conditionally setting the debounce value.

src/select/common.ts
```diff
@@ -37,10 +39,11 @@ export const useTypeahead = ({
 
   const handleOnKeyPress = (e: React.KeyboardEvent) => {
     e.persist();
-    // skip the enter key
-    if (e.key === "Enter") return;
+    // skip all the keys eg: Enter, Alt, Shift
+    if (e.key.length > 1) return;
+
```
anuraghazra, line 44: onKeyPress isn't working with SelectDropdown so i'm using the onKeyDown event handler, but onkeydown event handler also captures Enter, Shift, Alt, etc etc key strokes so i'm skipping the keys which are longer than 1 character, this ensure we are not concatenating the `typed` string with Shift, Enter & other keywords.

src/select/SelectDropdown.ts
navin-moorthy: We can remove the debounce, since we are not going to use it in multiselect

navin-moorthy:
Kindly change the file names of `useSelectState` to `SelectState` and `common` to `__utils`.
Just to be consistent with Reakit's structure, to have the familiar structure

src/select/__keys.ts
```diff
@@ -79,6 +79,10 @@ export const SELECT_KEYS = [
   "setSelected",
   "_setSelected",
   "isPlaceholder",
+  "inputValue",
+  "setInputValue",
+  "isCombobox",
+  "values",
   ...POPOVER_STATE_KEYS,
 ] as const;
```
navin-moorthy, line 87:
```
const COMPOSITE_STATE_KEYS = [
  "baseId",
  "unstable_idCountRef",
  "setBaseId",
  "unstable_virtual",
  "rtl",
  "orientation",
  "items",
  "groups",
  "currentId",
  "loop",
  "wrap",
  "unstable_moves",
  "unstable_angular",
  "unstable_hasActiveWidget",
  "registerItem",
  "unregisterItem",
  "registerGroup",
  "unregisterGroup",
  "move",
  "next",
  "previous",
  "up",
  "down",
  "first",
  "last",
  "sort",
  "unstable_setVirtual",
  "setRTL",
  "setOrientation",
  "setCurrentId",
  "setLoop",
  "setWrap",
  "reset",
  "unstable_setHasActiveWidget",
] as const;

const POPOVER_STATE_KEYS = [
  "baseId",
  "unstable_idCountRef",
  "visible",
  "animated",
  "animating",
  "setBaseId",
  "show",
  "hide",
  "toggle",
  "setVisible",
  "setAnimated",
  "stopAnimation",
  "modal",
  "unstable_disclosureRef",
  "setModal",
  "unstable_referenceRef",
  "unstable_popoverRef",
  "unstable_arrowRef",
  "unstable_popoverStyles",
  "unstable_arrowStyles",
  "unstable_originalPlacement",
  "unstable_update",
  "placement",
  "place",
] as const;

export const SELECT_STATE_KEYS = [
  ...COMPOSITE_STATE_KEYS,
  ...POPOVER_STATE_KEYS,
  "isPlaceholder",
  "isCombobox",
  "allowMultiselect",
  "selected",
  "setSelected",
  "removeSelected",
  "inputValue",
  "setInputValue",
  "values",
] as const;

export const SELECT_KEYS = SELECT_STATE_KEYS;
```

navin-moorthy (changes requested):
Some of the changes requesting !!
It's taking a little time.

Just to get it perfect.

src/select/common.ts
```diff
@@ -1,5 +1,65 @@
...
+import { SelectStateReturn } from "./useSelectState";
+
 export const enterOrSpace = (e: any, fn: Function) => {
   if (e.key === "Enter" || e.key === " ") {
     fn && fn(e);
   }
 };
+
+export function usePortalShortcut(
+  pickerListBoxRef: React.RefObject<HTMLElement>,
```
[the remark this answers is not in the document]
anuraghazra: Or lets just use a generic name, ref

navin-moorthy:
I just moved to `Ready for Review` since I already started 😛

src/select/Select.ts
```diff
@@ -0,0 +1,34 @@
...
+  useProps({ selected, visible, onChange }, { ...htmlProps }) {
+    React.useEffect(() => {
+      onChange && onChange(selected);
+    }, [selected]);
+
+    return {
+      role: "button",
+      "aria-expanded": visible,
+      "aria-haspopup": "listbox",
+      ...htmlProps,
```
navin-moorthy, line 25: I think, this can be moved to `SelectTrigger`,
`Select` is just a wrapper here

src/select/Select.ts
```diff
@@ -0,0 +1,34 @@
...
+import { createHook, createComponent } from "reakit-system";
+import { SELECT_KEYS } from "./__keys";
+
+export type SelectOptions = Pick<SelectStateReturn, "selected" | "visible"> & {
+  onChange?: (value: any) => void;
+};
+
+const useSelect = createHook<SelectOptions, BoxHTMLProps>({
+  name: "Select",
+  keys: SELECT_KEYS,
```
navin-moorthy, line 14:
```suggestion
  compose: useBox,
  keys: SELECT_KEYS,
```

src/select/Select.ts
navin-moorthy: It's already there in `SelectTrigger`, you can just remove these lines

anuraghazra: yeah SelectTrigger has that, just  have to remove this redundant code

src/select/__utils.ts
```diff
@@ -39,7 +32,7 @@ export function usePortalShortcut(
   });
 
   React.useEffect(() => {
-    const pickerListBox = pickerListBoxRef.current;
+    const pickerListBox = ref.current;
```
navin-moorthy, line 35:
```suggestion
    const selectMenu = ref.current;
```

`I used `pickerListBox` since that was my component, or you can give a generic name `menu`

anuraghazra: element?

src/select/SelectOption.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export type SelectItemHTMLProp = CompositeItemHTMLProps;
+
+const useSelectOption = createHook<SelectItemOptions, SelectItemHTMLProp>({
+  name: "SelectOption",
+  compose: useCompositeItem,
+  keys: ["value", ...SELECT_KEYS],
+  useProps(
+    { selected, value, setSelected, disabled },
+    { ref: htmlRef, onMouseMove: htmlOnMouseMove, ...htmlProps },
```
navin-moorthy, line 25: No need to extract ref anymore

```suggestion
    {  onMouseMove: htmlOnMouseMove, ...htmlProps },
```

src/select/SelectOption.ts
```diff
@@ -0,0 +1,61 @@
...
+    return {
+      role: "option",
+      "aria-label": value,
+      "data-selected": selected.includes(value),
+      "data-value": value,
+      onMouseMove,
+      onClick: e => {
+        e.preventDefault();
+        setSelected(value);
+      },
```
navin-moorthy, line 51: Kindly make the `onClick` handler more Reakit's way,

Get the `htmlOnClick`, live ref and call it inside this `onClick`.

This way we don't miss the user `onClick` also, if we follow the current way, 

Try passing an `onClick` on `SelectOption`, our `onClick` will not be executed. since it will be overriden.

Just follow the lead fo above `onMouseMove` function.

src/select/SelectOption.ts
```diff
@@ -0,0 +1,61 @@
...
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+      [disabled],
+    );
+
+    return {
+      role: "option",
+      "aria-label": value,
+      "data-selected": selected.includes(value),
+      "data-value": value,
```
navin-moorthy, line 46: I don't know this, can't we use the `value` directly as a attribute instead of `data-value`

I saw `Downshift` they use `id` 
Is there a reason, kindly let me know.

src/select/Select.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  useProps({ selected, onChange }, { ...htmlProps }) {
+    React.useEffect(() => {
+      onChange && onChange(selected);
+    }, [selected]);
+
+    return {
+      ...htmlProps,
+    };
```
navin-moorthy, line 23:
```suggestion
    return htmlProps;
```

src/select/Select.ts
```diff
@@ -0,0 +1,31 @@
...
+  onChange?: (value: any) => void;
+};
+
+const useSelect = createHook<SelectOptions, BoxHTMLProps>({
+  name: "Select",
+  keys: SELECT_KEYS,
+
+  useProps({ selected, onChange }, { ...htmlProps }) {
+    React.useEffect(() => {
+      onChange && onChange(selected);
```
navin-moorthy, line 18:
```suggestion
      onChange?.(selected);
```

src/select/SelectOption.ts
anuraghazra: I can probably use value attribute but if we use value attribute it would have a div with an attribute of value which i don't think is valid html.

src/select/__utils.ts
navin-moorthy: You copied the code from the `picker-v1` branch and there I used that variable name.

I am just asking for a better naming this component, or a generic name.

src/select/SelectOption.ts
navin-moorthy: Okay, okay, correct, only `<option>` can take `value` I believe.

Great !!

src/select/__utils.ts
anuraghazra: no no i'm saying how about the variable name "element" more generic

src/select/SelectTrigger.ts
```diff
@@ -1,27 +1,112 @@
...
+  compose: [usePopoverDisclosure],
   keys: SELECT_KEYS,
 
-  useProps({ unstable_referenceRef, visible }, { ref: htmlRef, ...htmlProps }) {
+  useProps(
+    options,
+    {
+      ref: htmlRef,
+      onClick: htmlOnClick,
+      onMouseDown: htmlOnMouseDown,
```
navin-moorthy, line 33: No more needed !!

anuraghazra: We do have Eslint setup right? Why isn't mine working? lol

src/select/SelectTrigger.ts
```diff
@@ -1,27 +1,112 @@
...
+        };
+      })(),
+    });
+
     return {
-      ref: useForkRef(unstable_referenceRef, htmlRef),
+      ref: useForkRef(ref, htmlRef),
+      onKeyDown,
+      tabIndex: 0,
       role: "button",
```
navin-moorthy, line 107:
```suggestion
```

Not needed since, `usePopoverDisclosure` uses `useDialogDisclosure` uses `useDisclosure` uses `useButton` 😛

src/select/SelectTrigger.ts
```diff
@@ -1,27 +1,112 @@
...
       "aria-haspopup": "listbox",
-      "aria-expanded": visible,
-      tabIndex: 0,
+      "aria-expanded": options.visible,
```
navin-moorthy, line 109: Also not needed, you can remove and check the attributes.

src/select/SelectTrigger.ts
anuraghazra: > usePopoverDisclosure uses useDialogDisclosure uses useDisclosure uses useButton

Brand new tongue twister.

src/select/__utils.ts
navin-moorthy: Oh !!!!!! 😆 

Absolutely fine

src/select/SelectTrigger.ts
```diff
@@ -1,27 +1,100 @@
...
+  | "selected"
+  | "setSelected"
+  | "show"
+  | "hide"
+  | "first"
+  | "last"
+> & { disabled?: boolean };
 
 const useSelectTrigger = createHook<SelectTriggerOptions, BoxHTMLProps>({
   name: "selectTrigger",
```
navin-moorthy, line 24:
```suggestion
  name: "SelectTrigger",
```

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,56 @@
...
+  };
 
 const useSelectMenu = createHook<SelectMenuOptions, BoxHTMLProps>({
   name: "SelectMenu",
-  keys: SELECT_KEYS,
+  compose: [useComposite, usePopover],
+  keys: ["maxHeight", ...SELECT_KEYS],
+  useOptions({ modal = false, ...options }) {
+    return { modal, ...options };
+  },
```
navin-moorthy, line 27: Any particular reason, why you have set `modal` to false.

No used inside `useProps` too. Just trying to understand.

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,56 @@
...
-      onKeyPress: handleOnKeyPress,
+      ref: useForkRef(ref, htmlRef),
+      role: "listbox",
+      "aria-orientation": "vertical",
+      "aria-hidden": !visible,
+      style: {
+        maxHeight: maxHeight,
+        overflowY: "scroll",
+        ...htmlStyle,
+      },
```
navin-moorthy, line 53: Do you think, we need to set these styles.
Instead of the user handling this?
If they want a scrolling container, they can set an maxHeight and the overflowY themselves right !!

anuraghazra: Okay will remove it. then we also need the maxHeight option

src/select/SelectMenu.ts
anuraghazra: Seen it here, so i put that in, https://github.com/reakit/reakit/blob/master/packages/reakit/src/Popover/Popover.ts#L25

navin-moorthy:
image

Can we have `aria-labelledby` for both `Trigger` and `Menu` - Design Pattern has it and Downshift has it.
But that is only needed when there's label to the Select.

How can we do this 🤔 
Do you think, we can provide a documention to provide `aria-labelledby` when using with a `label` ?

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,56 @@
...
     return {
-      role: "button",
-      "aria-expanded": visible,
-      "aria-haspopup": "listbox",
-      onKeyDown,
-      onKeyPress: handleOnKeyPress,
+      ref: useForkRef(ref, htmlRef),
+      role: "listbox",
+      "aria-orientation": "vertical",
+      "aria-hidden": !visible,
```
navin-moorthy, line 48: Why do you think, these are necessary?

I did not see these in `design pattern wai aria`

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,56 @@
...
+export type SelectMenuOptions = CompositeProps &
+  Pick<
+    SelectStateReturn,
+    "setSelected" | "visible" | "values" | "currentId" | "move"
+  > & {
+    maxHeight?: string | number;
+    modal?: boolean;
+  };
 
 const useSelectMenu = createHook<SelectMenuOptions, BoxHTMLProps>({
```
navin-moorthy, line 21:
```suggestion
export type SelectMenuHTMLProps = CompositeHTMLProps & PopoverHTMLProps;

const useSelectMenu = createHook<SelectMenuOptions, SelectMenuHTMLProps>({
```

src/select/SelectTrigger.ts
```diff
@@ -1,27 +1,100 @@
...
+  | "values"
+  | "selected"
+  | "setSelected"
+  | "show"
+  | "hide"
+  | "first"
+  | "last"
+> & { disabled?: boolean };
 
 const useSelectTrigger = createHook<SelectTriggerOptions, BoxHTMLProps>({
```
navin-moorthy, line 23:
```suggestion

const useSelectTrigger = createHook<
  SelectTriggerOptions,
  PopoverDisclosureHTMLProps
>({
```

src/select/SelectMenu.ts
navin-moorthy: If it's already there, then you don't have to put in, that will also be executed.
So no worries

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,54 @@
...
-        setTyped("");
-        keyClear.current = null;
-      }, 800);
-    };
+    usePortalShortcut(ref, { values, currentId, move });
 
     React.useEffect(() => {
-      if (typed !== "") {
-        clearKeyStrokes();
+      if (values) {
```
navin-moorthy, line 42:
```suggestion
    if (values && !selected?.length) {
```

Do this additional check so that the select doesn't focus the first option, when the value is selected

navin-moorthy:
Kindly check all the useEffect by removing the component wide `/* eslint-disable react-hooks/exhaustive-deps */` comment and add the deps appropriately.

src/select/SelectMenu.ts
anuraghazra: I don't know what do you meant by this but adding this check prevented the original behaviour

src/select/SelectOption.ts
```diff
@@ -0,0 +1,66 @@
...
+    const onClick = (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+      onClickRef.current?.(event);
+      if (event.defaultPrevented) return;
+      event.preventDefault();
+      setSelected(value);
+    };
+
+    return {
+      role: "option",
+      "aria-label": value,
```
navin-moorthy, line 52: Why do we need `aria-label` 🤔

src/select/SelectMenu.ts
navin-moorthy: When it is multiselect, it can focus the first option, itself. 🤔

src/select/SelectTrigger.ts
```diff
@@ -1,27 +1,107 @@
...
->;
+  | "visible"
+  | "values"
+  | "selected"
+  | "setSelected"
+  | "show"
+  | "hide"
+  | "first"
+  | "last"
+> & { disabled?: boolean };
```
navin-moorthy, line 23:
```suggestion

export type SelectTriggerOptions = PopoverDisclosureOptions &
  Pick<
    SelectStateReturn,
    | "visible"
    | "values"
    | "selected"
    | "setSelected"
    | "show"
    | "hide"
    | "first"
    | "last"
  >;
```

src/select/SelectMenu.ts
anuraghazra: I think we can do this check

```js
if (values && !selected?.length) {
        const firstItem = values[0];
        firstItem && move?.(firstItem.id);
      } else {
        const item = values.find(i => {
          return i.value === selected[selected.length - 1];
        })?.id;
        move?.(item as string);
      }
```

On Multiselect the focus would start from the last selected item

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,54 @@
...
 
     return {
-      role: "button",
-      "aria-expanded": visible,
-      "aria-haspopup": "listbox",
-      onKeyDown,
-      onKeyPress: handleOnKeyPress,
+      ref: useForkRef(ref, htmlRef),
+      role: "listbox",
+      style: htmlStyle,
```
navin-moorthy, line 51: you can remove this completely !!

anuraghazra: Which one?

navin-moorthy: Oops, selected everything 🤦 
the `style`

src/select/SelectMenu.ts
navin-moorthy: Great !!!

src/select/SelectMenu.ts
anuraghazra: Does not seem to be working without htmlStyle

navin-moorthy: You need to remove the destructuring above too.

navin-moorthy: If you have doubt, can I push a commit here 🤔

navin-moorthy: image

Don't desctructure and just let it pass, that's what I said.

anuraghazra: Ohhhh okay.

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,59 @@
...
-> & {
-  onChange?: (value: any) => void;
-};
+export type SelectMenuOptions = CompositeProps &
+  Pick<
+    SelectStateReturn,
+    "setSelected" | "visible" | "values" | "currentId" | "move" | "selected"
+  > & {
+    modal?: boolean;
+  };
```
navin-moorthy, line 21:
```suggestion
export type SelectMenuOptions = CompositeProps &
  PopoverProps &
  Pick<
    SelectStateReturn,
    "setSelected" | "visible" | "values" | "currentId" | "move" | "selected"
  >;
```

src/select/SelectMenu.ts
navin-moorthy: You can remove these.

src/select/SelectOption.ts
```diff
@@ -0,0 +1,64 @@
...
+      },
+      [disabled, onMouseMoveRef],
+    );
+
+    const onClick = (event: React.MouseEvent<HTMLElement, MouseEvent>) => {
+      onClickRef.current?.(event);
+      if (event.defaultPrevented) return;
+      event.preventDefault();
+      setSelected(value);
+    };
```
navin-moorthy, line 47: UseCallback here !!

src/select/SelectState.ts
```diff
@@ -0,0 +1,138 @@
...
+  const [selectedValue, setSelectedValue] = React.useState<string[]>(() => {
+    if (initialValue) {
+      return [initialValue];
+    }
+    return [];
+  });
+
+  const removeSelected = (value: string) => {
+    setSelectedValue(selectedValue.filter(item => item !== value));
+  };
```
navin-moorthy, line 88: useCallback here !!

src/select/SelectInput.ts
```diff
@@ -0,0 +1,52 @@
...
+      type: "text",
+      value: inputValue,
+      onKeyDown: e => {
+        if (e.key === "Backspace" && inputValue === "") {
+          removeSelected(selected[selected.length - 1]);
+        }
+      },
+      onChange: (e: React.ChangeEvent<HTMLInputElement>) => {
+        setInputValue(e.target.value);
+      },
```
navin-moorthy, line 37: If you are not experimenting, make them perfect as Reakit's way.

get htmlOnChange and htmlKeyDown, prevent default and disabled handle everything.

src/select/SelectInput.ts
```diff
@@ -0,0 +1,52 @@
...
+  | "inputValue"
+  | "setInputValue"
+  | "removeSelected"
+  | "selected"
+>;
+
+const useSelectInput = createHook<SelectInputOptions, BoxHTMLProps>({
+  name: "SelectInput",
+  compose: [useComposite, usePopoverDisclosure],
+  keys: SELECT_KEYS,
```
navin-moorthy, line 21: Proper types too !!!

navin-moorthy:
> image
> 
> Can we have `aria-labelledby` for both `Trigger` and `Menu` - Design Pattern has it and Downshift has it.
> But that is only needed when there's label to the Select.
> 
> How can we do this 🤔
> Do you think, we can provide a documention to provide `aria-labelledby` when using with a `label` ?

What do you think about this?

anuraghazra:
> What do you think about this?

I'll look into it to see if i can get this to work.

src/select/SelectMenu.ts
```diff
@@ -1,77 +1,67 @@
...
-const useSelectMenu = createHook<SelectMenuOptions, BoxHTMLProps>({
+export type SelectMenuHTMLProps = CompositeHTMLProps & PopoverHTMLProps;
+
+const useSelectMenu = createHook<SelectMenuOptions, SelectMenuHTMLProps>({
   name: "SelectMenu",
+  compose: [useComposite, usePopover],
   keys: SELECT_KEYS,
+  useOptions({ modal = false, ...options }) {
+    return { modal, ...options };
+  },
```
navin-moorthy, line 36: Just remove this line.

Everything else is good !!

navin-moorthy (changes requested):
LGTM !!

Make that one change and Merge the PR.

Input Combobox has some bugs, but that can be a seperate PR concentrating that.

navin-moorthy:
Also, fix the conflict before merging.

navin-moorthy (approved):

PR #12, opened 2020-08-28
Picker v1

navin-moorthy:
Single Picker


src/picker/PickerListBox.ts
```diff
@@ -0,0 +1,60 @@
...
+        const selectedValue = options.values.find(
+          value => value.value === options.selectedValue,
+        );
+
+        if (selectedValue?.id) {
+          options.move?.(selectedValue.id);
+        }
+      }
+      // eslint-disable-next-line react-hooks/exhaustive-deps
+    }, [options.selectedValue, options.values, options.move, options.visible]);
```
anuraghazra, line 46: Not sure but this bit of code was causing a bug in our Select's Multiselect behaviour it was causing the focus to reset to the first item select everytime.

navin-moorthy: But this code, just focuses the selected value when there is an selected value.
You can change the conditions, if you want to do the checks for your case.

PR #14, opened 2020-09-01
feat: update accordion with more features

navin-moorthy:


src/accordion/AccordionState.ts
```diff
@@ -147,7 +187,55 @@ function reducer(
...
+      return { ...state };
+    }
+
+    case "last": {
+      const total = buttons.length;
+
+      const lastItem = buttons[total - 1];
+      lastItem.ref?.current?.focus();
+
+      return { ...state };
```
anuraghazra, line 238: Seems like we can also utilize a helper function here to remove the redundancy.

src/accordion/AccordionItem.ts
```diff
@@ -19,20 +19,19 @@ export const useAccordionItem = createHook<
   AccordionItemHTMLProps
 >({
   name: "AccordionItem",
-  compose: [unstable_useId],
+  compose: unstable_useId,
   keys: ACCORDION_KEYS,
 
   useProps(options, { ref: htmlRef, ...htmlProps }) {
+    const { id, registerItem } = options;
```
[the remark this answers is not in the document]
navin-moorthy: I was just following the reakit codebase style, I have never seen any destructuring in options.

So I did not either.

I will just destructure the state options and leave it's own component options.

What do you think, we can do, destructure of directly in the arguments itself?

src/accordion/AccordionState.ts
navin-moorthy: Yup, I thought that too.

I just procrastinated it yesterday 😛 because of burn out(not in the mood)
I will do it today.

anuraghazra: Okay no worries, not high priority you can also do this on a different PR.

src/accordion/AccordionItem.ts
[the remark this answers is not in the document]
anuraghazra: Well i think for larger options lists we should opt for destructuring on the function body instead of argument list directly.

and for smaller option list we can just directly destructure everything.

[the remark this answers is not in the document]
navin-moorthy: Okay, acceptable. we both can follow the same. Thanks !!

src/accordion/AccordionState.ts
navin-moorthy: I have made some refactoring by removing the logic from  reducer and add it to the hook directly.

Let me know, if any more can be done.

src/accordion/AccordionState.ts
```diff
@@ -102,18 +136,10 @@ export function useAccordionState(
...
-    }, []),
-    prev: React.useCallback(id => {
-      dispatch({ type: "prev", id });
-    }, []),
-    first: React.useCallback(() => {
-      dispatch({ type: "first" });
-    }, []),
-    last: React.useCallback(() => {
-      dispatch({ type: "last" });
-    }, []),
```
anuraghazra: Yeah nice! it was really ugly before.

src/accordion/AccordionState.ts
anuraghazra: Yeah looks nice!

anuraghazra:
LGTM!

PR #15, opened 2020-09-02
feat(slider): ✨  add SliderInput and refactors

navin-moorthy:


src/slider/SliderFilledTrack.ts
```diff
@@ -4,21 +4,23 @@
...
   keys: SLIDER_FILLED_TRACK_KEYS,
-  compose: [useBox],
 
   useProps(options, { style: htmlStyle, ...htmlProps }) {
+    const { styles } = options;
+
     return {
-      style: { ...options.styles.innerTrackStyle, ...htmlStyle },
       ...htmlProps,
+      style: { ...htmlStyle, ...styles.innerTrackStyle },
```
anuraghazra, line 23: Changed the order to enforce the applied styles?

src/slider/SliderTrack.ts
```diff
@@ -4,35 +4,37 @@
...
 import { SLIDER_TRACK_KEYS } from "./__keys";
+import { UseSliderReturn } from "./SliderState";
 
 export type useSliderTrackOptions = UseSliderReturn & {
   /**
-   * The base `id` to use for the sliderTrack and it's components
+   * The base `id` to use for the sliderTrack
    */
   id?: string;
 };
```
anuraghazra, line 20: I guess we should pick only the types which the component is using. 

```ts
export type useSliderTrackOptions = Pick<
  UseSliderReturn,
  "refs" | "state" | "styles"
> & {
  /**
   * The base `id` to use for the sliderTrack
   */
  id?: string;
};
```

anuraghazra: Same with all other components.

src/slider/SliderTrack.ts
```diff
@@ -4,35 +4,37 @@
...
     return {
-      id,
-      ref: mergeRefs(htmlRef, options.refs.trackRef),
-      "data-disabled": dataAttr(options.state.isDisabled),
-      style: { ...options.styles.trackStyle, ...htmlStyle },
       ...htmlProps,
+      id,
+      "data-disabled": dataAttr(state.isDisabled),
+      style: { ...htmlStyle, ...styles.trackStyle },
+      ref: mergeRefs(htmlRef, refs.trackRef),
```
anuraghazra, line 37: We can maybe use Reakit's useForkRef too right? i saw the code it's very similar just that merge refs can take multiple arguments and useForkRef takes just two. 

*(it won't matter but jut to reduce duplicate utils & save few extra bytes haha)*

navin-moorthy: Yup, I can totally do it. No issues. Just a matter of preference.

I thought since we are using it from Chakra, so did not mean to change it.

src/slider/SliderFilledTrack.ts
navin-moorthy: I wantedly changed the order, because these styles are necessary so the users don't override it with thier styles. 

What do you think?

anuraghazra: Yup thats totally fine, i'm just curious.

src/slider/SliderTrack.ts
anuraghazra: Okay! although if you change it make sure it works as expected. :D

navin-moorthy: Yup, it works as it is, both has the same logic behind.

I have experimented with both of them.

src/slider/SliderTrack.ts
navin-moorthy: What is the advantage or use in picking the variables only we use in a particular component.

I just dont know, can you help me understand.

I think it as since we are spreading all the values from state, I just don't pick it.

anuraghazra: Yes we don't see any issues because we are spreading the state directly to the components, but i think it's better to be explicit about types it would help end users understand which component takes which props (even tho they would not care about it) & also would help us reduce confusion in the long run.

Yeah i know picking the types is a bit redundant and a bit of a chore but imo it's best we pick them.

For example:

the <Slider /> does not take the action, handler props but if we directly set the `UseSliderReturn` those props also shows as required props. 

Without Pick:
slider-props-1

With Pick:
slider-props-2


Do this on your free time, it's not high priority, or let me know if can also fix them if you want. :D

navin-moorthy: No Worries, I just learned this, I will do it myself, Thanks much for deeps explanation 🥇

anuraghazra (approved):

src/slider/SliderTrack.ts
anuraghazra: Awesome!

anuraghazra:
LGTM!

PR #21, opened 2020-09-08
feat(link): ✨  add breadcrumbs with link

navin-moorthy:
Added Reakit Link Component in favour of Breadcrumb.
Also added the Breadcrumb component - made easy because of a Link component.


src/breadcrumbs/stories/Breadcrumbs.stories.tsx
```diff
@@ -0,0 +1,74 @@
...
+export const Default = () => {
+  return (
+    <Breadcrumbs className="breadcrumb">
+      <ol>
+        <li>
+          <BreadcrumbLink href="https://www.w3.org/TR/wai-aria-practices-1.1/">
+            WAI-ARIA Authoring Practices 1.1
+          </BreadcrumbLink>
+        </li>
+        <li>
```
anuraghazra, line 21: I think the API should look like this, because users should not care about the markup nor how to use regular breadcrumbs in semantic html manner.

```js
<Breadcrumb>
  <BreadcrumbItem>
    <BreadcrumbLink href="#">Home</BreadcrumbLink>
  </BreadcrumbItem>

  <BreadcrumbItem>
    <BreadcrumbLink href="#">Docs</BreadcrumbLink>
  </BreadcrumbItem>

  <BreadcrumbItem isCurrentPage>
    <BreadcrumbLink href="#">Breadcrumb</BreadcrumbLink>
  </BreadcrumbItem>
</Breadcrumb>
```

navin-moorthy: I accepts what you are saying but,
How can we do this.

Primary rule of Reakit.
>  Each component returns a single HTML element

Take Radio as example,
https://reakit.io/docs/radio/

It's the users responsibility to provide a `label` around `Radio`
image

src/link/Link.ts
```diff
@@ -0,0 +1,66 @@
...
+      if (element.tagName !== "A") {
+        setRole("link");
+      }
+    }, []);
+
+    let externalLinkProps = {};
+
+    if (isExternal) {
+      externalLinkProps = { target: "_blank", rel: "noopener noreferrer" };
+    }
```
anuraghazra, line 51: we can directly do:

```js
const externalLinkProps = {
    ...isExternal && { target: "_blank", rel: "noopener noreferrer" }
 };
```

or just inline to

```js
    return {
      ref: useForkRef(ref, htmlRef),
      role,
      ...isExternal && { target: "_blank", rel: "noopener noreferrer" },
      ...htmlProps,
    };
```

src/breadcrumbs/stories/Breadcrumbs.stories.tsx
anuraghazra: Okay we can keep this, and approach the more declarative api when we will be building the designed components.

src/breadcrumbs/BreadcrumbLink.ts
```diff
@@ -0,0 +1,34 @@
+import { LinkHTMLProps, LinkOptions, useLink } from "../link";
+import { createComponent, createHook } from "reakit-system";
```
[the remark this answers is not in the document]
navin-moorthy: 😆 I notice this while reviewing you code but forgot on mine.
Fixed.

src/breadcrumbs/stories/Breadcrumbs.stories.tsx
navin-moorthy: Yup, we can remove the other parts when building a UI component.

src/link/Link.ts
navin-moorthy: Yup we could, just being explicit.
I will change it to inline one.

src/link/Link.ts
```diff
@@ -0,0 +1,66 @@
...
+  compose: useClickable,
+  keys: ["isExternal"],
+
+  useOptions(options) {
+    return { unstable_clickOnSpace: false, ...options };
+  },
+
+  useProps({ isExternal }, { ref: htmlRef, ...htmlProps }) {
+    const ref = React.useRef<HTMLElement>(null);
+    const [role, setRole] = React.useState<"link" | undefined>(undefined);
```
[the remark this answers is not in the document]
navin-moorthy: If they want link to behave as a button they can use `Button`

But, this is also possible, if you set `role="button"`, that will take the precedence.

Thats, not needed IMO, since we are not destructuring the role, when user gives a role, that will be returned.

You can achieve what you are asking with this snippet.
```
<Link as="button" role="button" onClick={() => alert("Custom Link")}>
      Reakit
    </Link>
```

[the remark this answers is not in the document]
anuraghazra: Yeah thats what i was saying, okay got it now!

anuraghazra (approved):

PR #23, opened 2020-09-08
Meter v1

navin-moorthy:
Added Meter.


src/meter/MeterState.ts
```diff
@@ -0,0 +1,9 @@
+import {
+  UseProgressProps,
+  useProgressState,
+  ProgressStateReturn,
+} from "../progress";
+
+export type UseMeterProps = UseProgressProps;
+export const useMeterState = useProgressState;
+export type MeterStateReturn = ProgressStateReturn;
```
navin-moorthy, line 9: Re-exporting the component and use them in storybook result in some warning.
I had to create a file and re export individually to remove those warnings.

navin-moorthy:
Storybook changes are some quick refactors.

src/meter/MeterState.ts
anuraghazra: Okay i guess we can left it as it is.

src/breadcrumbs/BreadcrumbLink.ts
```diff
@@ -1,6 +1,6 @@
 import { createComponent, createHook } from "reakit-system";
 
-import { LinkHTMLProps, LinkOptions, useLink } from "../link";
+import { LinkHTMLProps, LinkOptions, useLink } from "../Link";
```
anuraghazra, line 3: Linux systems does not care about case sensitive imports but windows does, so it was breaking the build.

anuraghazra:
Okay i think git messed up the Case sensitive imports, we have to rename the Link file with git mv command in order to fix it.

PR #28, opened 2020-09-09
Pagination v1

navin-moorthy:
Pagination Component


src/pagination/PaginationButton.ts
```diff
@@ -0,0 +1,106 @@
...
+            move?.(goto);
+          }
+          break;
+      }
+    }, [goto, next, prev, first, last, move]);
+
+    const ariaLabel =
+      getAriaLabel?.(goto, isCurrent) ?? isCurrent
+        ? `Page ${goto}`
+        : `Go to ${goto} Page`;
```
anuraghazra, line 91: Setting the label to `Goto {goto} page` screen readers can't pronounce the "prev" let's change it to "previous"

```js
    const ariaLabel =
      getAriaLabel?.(goto, isCurrent) ?? isCurrent
        ? `Page ${goto}`
        : `Go to ${goto === "prev" ? "previous" : goto} Page`;

```

src/pagination/PaginationButton.ts
```diff
@@ -0,0 +1,106 @@
+import { callAllHandlers, isNumber } from "@chakra-ui/utils";
+import React from "react";
+import { createComponent, createHook } from "reakit-system";
```
anuraghazra, line 3: Change order to `react` first :)

src/pagination/PaginationButton.ts
```diff
@@ -0,0 +1,106 @@
...
+    | "next"
+    | "prev"
+    | "first"
+    | "last"
+    | "isAtMax"
+    | "isAtMin"
+  > & {
+    goto: string | number;
+    getAriaLabel?: (goto: string | number, isCurrent: boolean) => string;
+  };
```
anuraghazra, line 23: Let's add better TS support for the goto prop

```js

export type TGoto = "next" | "prev" | "last" | "first" | number;

export type PaginationButtonOptions = ButtonOptions &
  Pick<
    PaginationStateReturn,
    | "currentPage"
    | "move"
    | "next"
    | "prev"
    | "first"
    | "last"
    | "isAtMax"
    | "isAtMin"
  > & {
    goto: TGoto;
    getAriaLabel?: (goto: TGoto, isCurrent: boolean) => string;
  };
```

src/pagination/stories/Pagination.stories.tsx
```diff
@@ -0,0 +1,83 @@
...
+        </li>
+        {state.pages.map(page => {
+          if (page === "start-ellipsis" || page === "end-ellipsis") {
+            return <li key={page}>...</li>;
+          }
+
+          return (
+            <li key={page}>
+              <PaginationButton
+                goto={page}
```
anuraghazra, line 37: NOTE!: we have to assert the page variable to `TGoto` 

```js
goto={page as TGoto}
```

src/pagination/PaginationButton.ts
```diff
@@ -0,0 +1,106 @@
...
+          last?.();
+          break;
+
+        default:
+          if (isNumber(goto)) {
+            move?.(goto);
+          }
+          break;
+      }
+    }, [goto, next, prev, first, last, move]);
```
anuraghazra, line 86: We can refactor this logic to this ->

```js
// prop destructure
{ currentPage, move, goto, getAriaLabel, ...options },

// ->
    const onClick = React.useCallback(() => {
      if (isNumber(goto)) {
        move?.(goto);
      } else {
        options[goto]?.();
      }
    }, [goto, move, options]);
```

anuraghazra (changes requested):

src/pagination/PaginationButton.ts
navin-moorthy: Thats done, I need to push big ordering every imports.
Waiting for your review to complete

src/pagination/PaginationButton.ts
navin-moorthy: Totally, more accurate, thanks 😄

src/pagination/PaginationButton.ts
navin-moorthy: 🤯 I did not think in that angle.

src/pagination/PaginationButton.ts
```diff
@@ -53,42 +55,22 @@ export const usePaginationButton = createHook<
     };
   },
 
-  useProps(
-    { currentPage, move, last, first, next, prev, goto, getAriaLabel },
-    { onClick: htmlOnClick, ...htmlProps },
-  ) {
+  useProps(options, { onClick: htmlOnClick, ...htmlProps }) {
```
anuraghazra, line 58: I think we should not just destructure it like so..

Because this way options would also contain all other functions and if users passes the same name to the "goto" prop the function would execute, although it's not a huge issue, since users would get typescript error if they pass anything else.

```js
options[goto]?.();
```

Let's do `{ currentPage, move, goto, getAriaLabel, ...options },` since this way the options would only contain the "next", "prev", "last", "first" methods.


Or else we can also double check the function names if you want to be safe

```js

  if (isNumber(goto)) {
    move?.(goto);
  } else if (["next", "prev", "last", "first"].includes(goto)) {
    options[goto]?.();
  }
```

anuraghazra:
Other than this small issue, everything LGTM!

src/pagination/PaginationButton.ts
navin-moorthy: I will go with the last one, much safer.

anuraghazra (approved):

PR #29, opened 2020-09-09
refactor: removed duplicate storybook code

anuraghazra:
Refactored duplicate code in progress/meter utils.js


src/utils.tsx
```diff
@@ -0,0 +1,262 @@
...
+export function useFakeProgression() {
+  const [value, setValue] = React.useState(0);
+
+  React.useEffect(() => {
+    const clearId = setInterval(() => {
+      setValue(prevValue => prevValue + 5);
+    }, 500);
+
+    if (value === 100) {
+      setValue(5);
```
navin-moorthy, line 86:
```suggestion
      clearInterval(clearId);
```

Add an argument like `type: "meter"` OR `type: "progress"`

navin-moorthy:
Really love this abstraction, saves a lot of line of duplication. ❤️ 
I did not give much importance since they were stories which may be changed later - so copy pasted it.
Thanks

src/utils.tsx
navin-moorthy: Because for Progress, we don't want to decrease the value back to 5.

navin-moorthy (approved):

PR #32, opened 2020-09-10
feat(meter): ✨  add meter functionalities

navin-moorthy:
Built an actual Meter 🚀

### DIFF
--- a/src/meter/MeterState.ts
+++ b/src/meter/MeterState.ts
@@ -43,6 +43,8 @@ export const useMeterState = (props: UseMeterProps = {}) => {
   high = handleInequalities(high, min, max);
   optimum = handleInequalities(optimum, min, max);
 
+  // More inequalities handled
+  //  low ≤ high (if both low and high are specified)
   if (low >= high) low = high;
   if (high <= low) high = low;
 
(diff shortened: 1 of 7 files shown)


src/meter/MeterState.ts
```diff
@@ -1,9 +1,99 @@
...
+
+  if (optimum >= min && optimum <= low) {
+    if (value >= min && value <= low) {
+      status = "safe";
+    } else if (value > low && value <= high) {
+      status = "caution";
+    } else {
+      status = "danger";
+    }
+  }
```
navin-moorthy, line 75: After understandin, see if you can minify this.

anuraghazra: Hmmmmm :thinking: 

We can extract the common logic tho.

```js

const isInRange = (value: number, min: number, max: number) => {
  return value >= min && value <= max;
};

const calculateStatus = (
  value: number,
  optimum: number,
  min: number,
  max: number,
  outRange: number,
) => {
  let status: TStatus = undefined;
  if (isInRange(optimum, min, max)) {
    if (value >= min && value <= max) {
      status = "safe";
    } else if (value < min && value >= outRange) {
      status = "caution";
    } else {
      status = "danger";
    }
  }

  return status;
};

  // especial case
  if (isInRange(optimum, low, high)) {
    if (value >= low && value <= high) {
      status = "safe";
    } else {
      status = "caution";
    }
  }

  status = calculateStatus(value, optimum, high, max, low);
  status = calculateStatus(value, optimum, min, low, high);
```

src/meter/MeterState.ts
```diff
@@ -1,9 +1,99 @@
...
+ * the default value of a slider by using mid-point
+ * between the min and the max.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/range
+ */
+export function getDefaultOptimumValue(min: number, max: number) {
+  return max < min ? min : min + (max - min) / 2;
+}
+
+export function handleInequalities(value: number, min: number, max: number) {
```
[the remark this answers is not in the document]
navin-moorthy: clamp looks good !!

src/meter/MeterState.ts
navin-moorthy:
```js
if (isInRange(optimum, min, max)) {
    if (value >= min && value <= max) {
      status = "safe";
    } else if (value < min && value >= outRange) {
      status = "caution";
    } else {
      status = "danger";
    }
  }
```

else if changes for two cases, I guess you did not notice. I have made changes and minified it.

src/meter/__utils.ts
```diff
@@ -0,0 +1,62 @@
...
+type CalculateStatusProps = {
+  value: number;
+  optimum: number;
+  min: number;
+  max: number;
+  low: number;
+  high: number;
+};
+
+export const calculateStatus = (props: CalculateStatusProps) => {
```
anuraghazra, line 36: Yup, looks good, i was also considering to extract the whole logic, sine it will help us unit test this part, this part really needs proper testing.

anuraghazra (approved):

src/meter/__utils.ts
navin-moorthy: Got it !!

PR #38, opened 2020-09-11
test: added tests for toast

anuraghazra:


package.json
```diff
@@ -81,6 +81,7 @@
     "jest-matcher-utils": "26.4.2",
     "lint-staged": "10.3.0",
     "prettier": "2.1.1",
+    "raf": "^3.4.1",
```
anuraghazra, line 84: have to remove it, it should work, but not sure why it isn't working, i had to approach manually.

PR #46, opened 2020-09-16
infra: added auto key generation script

anuraghazra:


navin-moorthy:
LGTM 🚀 

We needed this from beginning.

navin-moorthy (approved):

PR #47, opened 2020-09-16
feat(accordion): ✨  update accordion to expose panel open

navin-moorthy:
Update Accordion Component


src/accordion/stories/index.css
```diff
@@ -0,0 +1,109 @@
+.Accordion {
```
anuraghazra, line 1: Can we use small casing for css classes?

src/accordion/AccordionPanel.ts
```diff
@@ -26,30 +27,23 @@ export const useAccordionPanel = createHook<
...
-    const style = {
-      display: `${isOpen ? "block" : "none"}`,
-      ...htmlStyle,
-    };
+    const { item, isOpen } = useAccordionItemContext();
+    const buttonId = item?.button?.id;
 
     return {
       role: "region",
       "aria-labelledby": `${buttonId ? buttonId : undefined}`,
```
anuraghazra, line 44: It will be undefined if it's not found anyways right? so no need to check if it's undefined and then fallback to undefined.

src/accordion/stories/index.css
navin-moorthy: I just copy pasted from the WAI-ARIA site.
Since this is just an example, I did not care about the case.

src/accordion/AccordionPanel.ts
navin-moorthy: Oh, my bad. I changed it in Panel component but forgot here.
Point !!

src/accordion/stories/index.css
navin-moorthy: I have changed it !!

anuraghazra (approved):

PR #61, opened 2020-09-29
refactor(range-calendar): ♻️  review changes and fix bugs with space

navin-moorthy:
- Removed getCellDateProps() & moved everything to Calendar Cell
- Above change reduces lot of lines of code
- Fixed " " - OnKeyDown issue with preventDefault on top.
- Renamed CalendarRangeState to RangeCalendarState.
- Removed RangeCalendar because that is Calendar Grid === getCalendarBody(Aria's)
- Updated Controllable Storybook Examples
- Fixed styles for wrong data-attributes selectors
- Updated Announcer for normal Calendar as well


src/calendar-v1/CalendarGrid.ts
```diff
@@ -70,43 +73,91 @@ export const useCalendarGrid = createHook<
...
       calendarId,
+      setAnchorDate,
     } = options;
     const ref = useRef<HTMLElement>(null);
 
-    const onKeyDown = createOnKeyDown({
-      onKeyDown: htmlOnKeyDown,
-      preventDefault: true,
-      keyMap: (event: KeyboardEvent) => {
-        const shift = event.shiftKey;
```
anuraghazra, line 80: Any particular reasons for removing createOnKeyDown? 🤔

src/calendar-v1/CalendarCell.ts
```diff
@@ -36,23 +32,64 @@ export const useCalendarCell = createHook<
...
+      dateValue,
+      highlightDate,
+      highlightedRange,
+      isDisabled,
+      currentMonth,
+    } = options;
 
-    const onMouseEnter = () => {
-      highlightDate && highlightDate(date);
+    let calendarProps: any = {
```
anuraghazra, line 45: let's use `Record<string, any>` instead of `any`

src/calendar-v1/CalendarCell.ts
```diff
@@ -36,23 +32,64 @@ export const useCalendarCell = createHook<
...
+        : false;
+      const isSelectionEnd = highlightedRange
+        ? isSameDay(date, highlightedRange.end)
+        : false;
+
+      const onMouseEnter = () => {
+        if (isDisabled) return;
+
+        highlightDate?.(date);
+      };
```
anuraghazra, line 69: I kinda liked that this chunk of logic was in the state return `getCellOptions`, it's noisy in here. 🤔 maybe we can refactor/extract is somewhere.

navin-moorthy: But I thought it's not necessary to be in a state.
Because they are specific to the component.

May be we can refactor it.

anuraghazra: > But I thought it's not necessary to be in a state.
> Because they are specific to the component.
> 
> May be we can refactor it.

Yeah, although it was nicer since it was lazily in the state wasn't evaluated and has access to all the variables already since it had access to it's parent scope. just cleaner in general. we can maybe also extract it into another function if you don't want to put it on the state.

src/calendar-v1/CalendarGrid.ts
navin-moorthy: Nope, I was debugging " " issue and forgot to change it back.

src/calendar-v1/CalendarCell.ts
navin-moorthy: Yup, your points are vaild, one adv is it had access to all the variables.
I don't want them to be in the state because they don't belong there.
I have refactored it a bit, help me if you had a better idea.

navin-moorthy: Removing them, made the state cleaner

PR #69, opened 2020-10-05
fix: breadcrumb isCurrent disabled should be false

anuraghazra:
aria-current=page should not have be disabled, since it would prevent it from being tabbable and voice assistants won't announce it either.

image

https://www.w3.org/TR/wai-aria-practices-1.1/#breadcrumb


navin-moorthy (approved):
LGTM 🎉

PR #70, opened 2020-10-05
Datepicker v3

navin-moorthy:


src/datepicker/DatePicker.ts
```diff
@@ -1,4 +1,4 @@
-import { createOnKeyDown } from "reakit-utils";
+import { createOnKeyDown, isSelfTarget } from "reakit-utils";
```
anuraghazra, line 1: unused import

src/datepicker/DatePicker.ts
```diff
@@ -5,6 +5,7 @@ import { ariaAttr, callAllHandlers } from "@chakra-ui/utils";
 
 import { DATE_PICKER_KEYS } from "./__keys";
 import { DatePickerStateReturn } from "./DatePickerState";
+import { first } from "lodash";
```
anuraghazra: let's install lodash-es instead of lodash... it has lot of issues with treeshaking.

src/calendar/stories/Calendar.stories.tsx
```diff
@@ -3,22 +3,16 @@ import { Meta } from "@storybook/react";
 import { addDays, addWeeks, subWeeks } from "date-fns";
 
 import "./index.css";
+import { DateValue } from "../index";
```
anuraghazra, line 6: import from "index.d" just to be specific that we are importing from the d.ts file

src/datepicker/DatePicker.ts
```diff
@@ -1,22 +1,104 @@
...
+
+          return {
+            ArrowDown: () => {
+              event.preventDefault();
+              event.stopPropagation();
+              isAlt && show();
+            },
+          };
+        },
+      });
```
anuraghazra, line 80: We can do 

```js
      const onKeyDown = createOnKeyDown({
        onKey: htmlOnKeyDown,
        preventDefault: true,
        keyMap: event => {
          const isAlt = event.altKey;

          return {
            ArrowDown: () => {
              isAlt && show();
            },
          };
        },
      });
```

src/datepicker/DatePickerState.ts
```diff
@@ -1,53 +1,105 @@
...
+    onChange,
+    shouldUpdate: (prev, next) => prev !== next,
   });
+  const dateValue = value != null ? new Date(value) : undefined;
 
+  // Intercept setValue to make sure the Time section is not changed by date selection in Calendar
+  const selectDate = (newValue: Date) => {
+    if (dateValue) {
+      setTime(newValue, dateValue);
+    }
```
anuraghazra, line 51: Fix the type by accepting DateValue instead of Date

```js
  const selectDate = (newValue: DateValue) => {
    if (dateValue) {
      setTime(new Date(newValue), dateValue);
    }

    setValue(newValue);
    popover.hide();
  };
```
And remove the @ts-ignore at line 69.

anuraghazra (changes requested):

src/datepicker/DatePicker.ts
navin-moorthy: This is not used! My VsCode auto imported I believe

src/datepicker/DatePickerState.ts
```diff
@@ -5,49 +5,64 @@
...
   const [value, setValue] = useControllableState({
-    value: initialValue,
+    value: initialDate,
     defaultValue,
     onChange,
     shouldUpdate: (prev, next) => prev !== next,
   });
-  const dateValue = value != null ? new Date(value) : undefined;
+
+  const dateValue = value && isValid(value) ? new Date(value) : undefined;
```
anuraghazra, line 52: Can we do 

```js
const dateValue = isValid(value) ? new Date(value) : undefined;
```

does the isValid handles the undefined argumant?

navin-moorthy: Nope, it doesn't.

Can we write our own isValid with datefn function?

anuraghazra: Yeah just enhance it, use their isValid function and check for undefined values

PR #71, opened 2020-10-05
feat: added segment component

anuraghazra:
Extracted and isolated date segment component to make it more generic and reusable. 

We will implement the Segment component to calendar component after merging #70 

And this PR should be merged to datepicker-v2 branch #58


src/segment-spinner/stories/index.css
```diff
@@ -0,0 +1,21 @@
...
+  outline: none;
+}
+
+pre {
+  font-size: 12px;
+  color: rgb(101, 100, 124);
+}
+pre ~ pre {
+  margin-top: 35px;
+}
```
navin-moorthy, line 21: Don't style any tag directly, that affects other story aswell.
Add a unique class and target directly.

src/segment-spinner/__utils.ts
```diff
@@ -0,0 +1,233 @@
+export type DateValue = string | number | Date;
+
```
navin-moorthy, line 2: Move the export down !!

src/segment-spinner/SegmentField.ts
```diff
@@ -0,0 +1,49 @@
...
+        return {
+          Tab: () => {
+            isShift ? options.previous() : options.next();
+          },
+        };
+      },
+    });
+
+    return { ...htmlProps, onKeyDown };
+  },
```
navin-moorthy, line 38: This is unnecessary here, since tab movement will happen by default !!

src/segment-spinner/Segment.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    /*
+      Haz: 
+      Ensure tabIndex={0} 
+      Tab is not the only thing that can move focus in web pages
+      For example, on iOS you can move between form elements using 
+      the arrows above the keyboard
+    */
+    return {
+      ...htmlProps,
```
navin-moorthy, line 56:
```suggestion
```

No need to pass htmlProps again.

`composite` already has the htmlProps.

src/segment-spinner/Segment.ts
```diff
@@ -0,0 +1,229 @@
...
+    isRequired?: boolean;
+  };
+
+export type SegmentHTMLProps = CompositeItemHTMLProps & BoxHTMLProps & DOMProps;
+
+export type SegmentProps = SegmentOptions & SegmentHTMLProps;
+
+export const useSegment = createHook<SegmentOptions, SegmentHTMLProps>({
+  name: "Segment",
+  compose: [useBox, useCompositeItem],
```
navin-moorthy, line 35: No need of `useBox`

src/segment-spinner/Segment.ts
```diff
@@ -0,0 +1,229 @@
...
+          break;
+        }
+      }
+    };
+
+    const onFocus = () => {
+      setEnteredKeys("");
+    };
+
+    const id = useId(htmlProps.id);
```
navin-moorthy, line 185: Use Reakit's useId

src/segment-spinner/Segment.ts
```diff
@@ -0,0 +1,229 @@
...
+      // Editable segment
+      default:
+        return mergeProps(spinButtonProps, {
+          id,
+          "aria-label": segment.type,
+          "aria-labelledby": `${state["aria-labelledby"]} ${id}`,
+          tabIndex: state.isDisabled ? undefined : 0,
+          onKeyDown,
+          onFocus,
+          onMouseDown: (e: MouseEvent) => e.stopPropagation(),
```
navin-moorthy, line 217: get the handler from htmlProps and combine them using CallAllHandlers

navin-moorthy:
See my DatePicker V3 branch for all the fixes directly I did yesterday.
We can merge this branch after merging my branch my and branch and cross-reference it.

src/segment-spinner/Segment.ts
anuraghazra: Does not seem to be working and causing infinite rerenders. the implementation of both useIds are different.

src/segment-spinner/SegmentState.ts
```diff
@@ -0,0 +1,164 @@
...
+const TYPE_MAPPING = {
+  dayperiod: "dayPeriod",
+};
+
+export interface SegmentStateProps {
+  value?: Date;
+  defaultValue?: Date;
+  formatOptions?:
+    | (Intl.DateTimeFormatOptions & { timeStyle?: string; dateStyle?: string })
+    | undefined;
```
navin-moorthy, line 52:
```suggestion
formatOptions?: Intl.DateTimeFormatOptions & {
    timeStyle?: string;
    dateStyle?: string;
  };
```

navin-moorthy: I believe `undefined` is not needed because `?:` of optional type

anuraghazra: Weird thing i did not noticed this

src/segment-spinner/Segment.ts
```diff
@@ -0,0 +1,229 @@
...
+          children: segment.text,
+          ...htmlProps,
+        });
+    }
```
navin-moorthy, line 221:
```suggestion
         return mergeProps(
          {
            ...spinButtonProps,
            id,
            "aria-label": segment.type,
            "aria-labelledby": `${state["aria-labelledby"]} ${id}`,
            tabIndex: state.isDisabled ? undefined : 0,
            onKeyDown,
            onFocus,
            onMouseDown: (e: MouseEvent) => e.stopPropagation(),
            children: segment.text,
          },
          htmlProps,
        );
```

navin-moorthy: This change gives better clarity on which props will be over-written

anuraghazra: This breaks the behaviour

navin-moorthy: Oh right right !! handlers are not merged in this case. We need to use another mergeProps which is unnecessary.
No change required here.

navin-moorthy: Just use the htmlProps handlers with callAllHandlers thats necessary.

PR #73, opened 2020-10-07
test: added tests for calendar

anuraghazra:


src/calendar/__tests__/Calendar.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+                  <CalendarCellButton {...state} date={day} />
+                </CalendarCell>
+              ))}
+            </tr>
+          ))}
+        </tbody>
+      </CalendarGrid>
+    </CalendarWrapper>
+  );
+};
```
navin-moorthy, line 70: Use the component from stories.

OR create a folder structure like meter component with __examples__ containing tests & stories based on a common component.

src/calendar/__tests__/Calendar.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+    );
+
+    expect(testId("current-year")).toHaveTextContent("October 2020");
+    press.Tab();
+    press.Tab();
+    press.Tab();
+    press.Tab();
+    press.Tab();
+
+    expect(label("Today, Wednesday, October 7, 2020 selected")).toHaveFocus();
```
navin-moorthy, line 118: Test fails today, because todays date changes.
Select a date that's not today.

src/calendar/__tests__/RangeCalendar.test.tsx
```diff
@@ -0,0 +1,170 @@
...
+    );
+
+    expect(testId("current-year")).toHaveTextContent("October 2020");
+    press.Tab();
+    press.Tab();
+    press.Tab();
+    press.Tab();
+    press.Tab();
+
+    expect(label("Today, Wednesday, October 7, 2020 selected")).toHaveFocus();
```
navin-moorthy, line 123: Same error like Calendar.

navin-moorthy (commented):
Just fix the above changes - Rest good to go.

src/calendar/__tests__/Calendar.test.tsx
anuraghazra: We don't want to reuse the stories component, I believe that test components should be isolated otherwise if we change something in the stories' component that could possibly break the tests.

src/calendar/__tests__/Calendar.test.tsx
anuraghazra: lol my bad, these test are designed to work only once per year. :p 😆

src/calendar/__tests__/Calendar.test.tsx
navin-moorthy: Okay, if you think so. Valid point !!

navin-moorthy (approved):
LGTM 🎉

PR #75, opened 2020-10-08
Timepicker v2

navin-moorthy:


src/timepicker/TimePickerState.ts
```diff
@@ -1,68 +1,106 @@
...
   const hourState = useTimePickerColumnState({
-    value: time!.getHours() % 12,
-    onChange: v => {
-      const d = new Date(time?.toISOString() as string);
-      d.setHours(v as number);
-      setTime(d);
-    },
-    onSelection: () => disclosure.hide(),
+    value: hour,
+    onChange: setHour,
```
anuraghazra, line 70: TS Error here

image


Because the function signature is different.

src/timepicker/TimePickerState.ts
```diff
@@ -1,68 +1,106 @@
...
+
+  const segmentState = useSegmentState({
+    value: time,
+    defaultValue,
+    onChange: setSegmentTime,
+    formatOptions,
+    placeholderDate,
+  });
+
+  const popover = usePopoverState({ modal: true, ...props });
```
anuraghazra, line 102: Move this to the beginning, since we are using popover before `onSelection: () => popover.hide(),` we are not getting error since it's wrapped in a function if we wanted to do `onSelection: popover.hide` it will throw error

src/timepicker/TimePickerState.ts
navin-moorthy: I am aware. Did that at the last moment yesterday.
I will fix it.

src/timepicker/stories/index.css
```diff
@@ -27,11 +54,23 @@
...
+  width: 5px;
+}
+
+.timepicker__content ::-webkit-scrollbar-track {
+  background-color: #dde0ff;
+}
+
+.timepicker__content ::-webkit-scrollbar-thumb {
+  background-color: #9097ff;
+}
```
anuraghazra, line 76: Add this CSS

```css

.timepicker__column--value[data-selected="true"] {
  background-color: #1e65fd;
  color: white;
}
```
it will show the selected time value..
image

src/timepicker/TimePickerState.ts
```diff
@@ -50,19 +50,19 @@ export const useTimePickerState = (props: TimePickerStateProps = {}) => {
...
+    setMinuteProp(minute as number);
+    setTime(hour, minute as number, meridian);
   };
 
-  const setMeridian = (meridianValue: string) => {
-    setMeridianProp(meridianValue);
-    setTime(hour, minute, meridianValue);
+  const setMeridian = (meridianValue: number | string) => {
+    setMeridianProp(meridianValue as string);
+    setTime(hour, minute, meridianValue as string);
```
anuraghazra: Hmmmm 🤔 this works but i think we can utilize discriminated union functions on the onChange type

```ts
  onChange?: ((v: number) => void) | ((v: string) => void);
```

navin-moorthy: I just fixed it temporarily. We need to fix it by providing better onChange types.

Need you help, I am not as good as you in TS.

anuraghazra: Yup thats what we want, better onChange, this should work.

onChange?: ((v: number) => void) | ((v: string) => void);

src/timepicker/TimePicker.ts
```diff
@@ -15,8 +29,62 @@ export const useTimePicker = createHook<TimePickerOptions, TimePickerHTMLProps>(
...
+        "aria-haspopup": "dialog",
+        "aria-expanded": visible,
+        "aria-owns": visible ? dialogId : undefined,
+        "aria-disabled": ariaAttr(isDisabled),
+        "aria-readonly": ariaAttr(isReadOnly),
+        onKeyDown: callAllHandlers(htmlOnKeyDown, onKeyDown),
+        onClick: callAllHandlers(htmlOnClick, onClick),
+        onMouseDown: callAllHandlers(htmlOnMouseDown, onMouseDown),
+        ...htmlProps,
+      };
```
anuraghazra, line 87: This logic is very similar to DatePicker 🤔 let's refactor it to a common component.

src/timepicker/TimePickerColumnState.ts
```diff
@@ -1,46 +1,25 @@
...
 import { useCompositeState } from "reakit";
+import { useControllableState } from "@chakra-ui/hooks";
 
-interface Props {
-  onChange?: (v: number | string) => void;
-  onSelection?: (v: number | string) => void;
-  value?: number | string;
+interface TimePickerColumnInitialProps {
+  value?: number;
+  onChange?: (v: number) => void;
```
anuraghazra, line 6: Why are we only accepting number here? 🤔 what happened to "AM" and "PM" meridians

src/timepicker/TimePickerColumnValue.ts
```diff
@@ -26,33 +35,39 @@ export const useTimePickerColumnValue = createHook<
...
+      setCurrentId,
+      move,
+      selected,
+      value,
+      id,
+      setSelected,
+      visible,
+    } = options;
+    const ref = React.useRef<HTMLElement>();
+    const valueProp = value === "AM" ? 0 : value === "PM" ? 1 : +value;
```
anuraghazra, line 52: Okay got it, so you are forcing number inputs and casting AM to 0 and PM to 1. 

The column still works with any other value right? 🤔  or not?

image

src/timepicker/TimePickerColumnValue.ts
```diff
@@ -1,22 +1,31 @@
...
-export type TimePickerColumnValueOptions = CompositeItemOptions &
-  Pick<TimePickerColumnStateReturn, "selected" | "move" | "onSelection"> & {
+export type TimePickerColumnValueOptions = ButtonOptions &
+  CompositeItemOptions &
+  Pick<
+    TimePickerColumnStateReturn,
+    "selected" | "move" | "setSelected" | "visible"
+  > & {
     value: number | string;
     visible?: boolean; // used to track mounting
```
anuraghazra, line 24: We can remove it since we are getting the visible prop from StateReturn

src/timepicker/TimePicker.ts
navin-moorthy: I am thinking the same, We need to extract this to a separate component as Picker and use it in other places.
I will do it tomorrow.

Similar Behaviours,
DatePicker - TimePicker
DatePickerTrigger - TimePickerTrigger
DatePickerContent - TimePickerContent.

They all work towards the same goal, makes combobox.

src/timepicker/TimePickerState.ts
```diff
@@ -1,68 +1,125 @@
...
 import { useTimePickerColumnState } from "./TimePickerColumnState";
 
-export interface TimePickerStateProps {
-  onChange?: (v: Date | null) => void;
-  visible?: boolean;
+export interface TimePickerStateProps extends PopoverInitialState, InputBase {
   value?: string;
+  defaultValue?: string;
+  onChange?: (v: string) => void;
+  formatOptions?: Intl.DateTimeFormatOptions;
```
anuraghazra, line 17: Add the timeStyle / dateStyle bindings

src/timepicker/TimePickerColumnState.ts
navin-moorthy: This happened 😛 

0 for "AM"  |    1 for "PM"

What do you think ?
 
image

src/timepicker/TimePickerState.ts
navin-moorthy: We need to add and make a type for ourselves.

I think we can better add `utils/type.ts` for common types among components like other libraries.

src/timepicker/TimePickerColumnValue.ts
navin-moorthy: So, what do you say.

This is a bad idea 🤔 , needs a correction or revert back to how it was ?

src/timepicker/TimePickerState.ts
anuraghazra: Yeah

src/timepicker/TimePicker.ts
anuraghazra: Yeah better we extract it... let's name it "PickerBase" Picker might be confusing or refer to a concrete component

src/timepicker/TimePickerColumnState.ts
anuraghazra: I'm fine with it, as long as the Column still accepts and works with arbitrary values

src/timepicker/TimePicker.ts
navin-moorthy: Got it !!

src/timepicker/TimePickerColumnValue.ts
anuraghazra: Let's try it out and see what problems it makes with arbitrary values

anuraghazra: the basic keyboard navigations works, but selection returns NaN

image
image

anuraghazra: If we use the map index then it works... so the behaviour changes to index based instead of value based.

image

src/utils/index.ts
```diff
@@ -27,3 +27,5 @@ export function isTouch() {
       window.navigator.msMaxTouchPoints > 0,
   );
 }
+
+export * from "./types";
```
anuraghazra, line 31: Let's not reexport the types here again.. i would like to be explicit about importing the types from "./utils/types"

navin-moorthy: Okay, that would be nice too. We know we are using the types. 👍

src/timepicker/stories/TimePicker.stories.tsx
```diff
@@ -1,87 +1,134 @@
...
+  TimePickerContent,
+  TimePickerStateProps,
+  TimePickerTrigger,
+  TimePickerSegment,
+  TimePickerSegmentField,
+  stringifyTime,
+} from "../index";
+import { addDays, addWeeks, subWeeks } from "date-fns";
+import { DateValue } from "../../utils/types";
+import { stringify } from "querystring";
```
anuraghazra, line 19: remove unused imports

navin-moorthy: My mistake !!

anuraghazra (approved):

anuraghazra:
Noice!

PR #84, opened 2020-10-12
fix: timepicker voiceover a11y

anuraghazra:


navin-moorthy (approved):
VO in mac works well.

LGTM 🎉

PR #85, opened 2020-10-12
feat(time-picker): ✨  add keyboard navigation to the column

navin-moorthy:
Adds keyboard arroa key navigation to Timepicker Column

### DIFF
--- a/src/timepicker/TimePickerContent.ts
+++ b/src/timepicker/TimePickerContent.ts
@@ -1,4 +1,7 @@
+import * as React from "react";
 import { createComponent, createHook } from "reakit-system";
+import { createOnKeyDown, useForkRef } from "reakit-utils";
+import { focus, getNextTabbable, getPreviousTabbable } from "@chakra-ui/utils";
 
 import {
   PickerBaseContentHTMLProps,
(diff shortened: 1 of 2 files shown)


src/timepicker/TimePickerContent.ts
```diff
@@ -21,6 +24,40 @@ export const useTimePickerContent = createHook<
...
+  useProps(_, { ref: htmlRef, onKeyDown: htmlOnKeyDown, ...htmlProps }) {
+    const ref = React.useRef<HTMLElement>(null);
+
+    const onKeyDown = React.useMemo(() => {
+      return createOnKeyDown({
+        onKeyDown: htmlOnKeyDown,
+        keyMap: () => {
+          return {
+            ArrowRight: () => {
+              if (ref.current) {
```
anuraghazra, line 37: I think we can put the `if (ref.current)` check outside ArrowRight & ArrowLeft. before the return

anuraghazra: Throws TS null assertion error.

navin-moorthy: Added Guard Clause instead as discussed.

anuraghazra (approved):

anuraghazra:
LGTM! 🎉

PR #86, opened 2020-10-12
test: added test for timepicker

anuraghazra:


src/timepicker/TimePickerColumnState.ts
```diff
@@ -8,7 +8,7 @@ export type ColumnType = "hour" | "minute" | "meridian";
 
 export interface TimePickerColumnInitialState extends ValueBase<Date> {
   visible?: boolean;
-  type?: ColumnType;
+  columnType?: ColumnType;
```
anuraghazra, line 11: Changed the `type` to `columnType` because jestdom is throwing erros with because native elements also have `type` prop.

navin-moorthy (approved):
LGTM 🎉

PR #88, opened 2020-10-12
refactor(pickers): ♻️  improve state from dateValue to string

navin-moorthy:
Improved DateValue handling to native string handling


src/datepicker/DatePickerState.ts
```diff
@@ -11,107 +11,95 @@ import {
...
+  const selectDate = (newValue: string) => {
+    const newDate = parseDate(newValue);
+    if (newDate) setValue(newDate);
     popover.hide();
   };
 
   const segmentState = useSegmentState({
-    value: dateValue,
-    defaultValue,
+    value,
```
anuraghazra, line 69: the value here is string, but useSegmentState only accepts Date 🤔

src/calendar/CalendarCellButton.ts
```diff
@@ -11,7 +11,7 @@ import { createComponent, createHook } from "reakit-system";
 import { ariaAttr, callAllHandlers } from "@chakra-ui/utils";
 import { ButtonHTMLProps, ButtonOptions, useButton } from "reakit";
 
-import { isInvalid } from "./__utils";
+import { isInvalidDate } from "../utils";
```
anuraghazra, line 14: let's change this function from isInvalidDate to isDateInValidRange or something like that 🤔

src/utils/date.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function isLeapYear(year: number) {
+  return (year % 4 == 0 && year % 100 != 0) || year % 400 == 0;
+}
+
+export function stringifyDate(date: Date) {
+  return `${date.getFullYear()}-${pad(date.getMonth() + 1)}-${pad(
+    date.getDate(),
+  )}`;
+}
```
anuraghazra, line 59: Hmm another way i found which seems more native 🤔 you can use that too maybe... 

```ts
export function stringifyDate(date: Date) {
  new Date(date.getTime() - date.getTimezoneOffset() * 60000)
    .toISOString()
    .split("T")[0];
}
```

anuraghazra: Oh wait... why we are not doing this? 

```ts
export function stringifyDate(date: Date) {
  return format(date, "yyyy-MM-dd");
}
```

src/utils/date.ts
```diff
@@ -0,0 +1,74 @@
...
+
+function getMaxDay(month: number, isLeapYear = false) {
+  const _31DaysMonth = [1, 3, 5, 7, 8, 10];
+  const _30DaysMonth = [4, 6, 9, 11];
+
+  if (_31DaysMonth.includes(month)) return 31;
+  if (_30DaysMonth.includes(month)) return 30;
+  if (month === 2 && isLeapYear) return 29;
+  return 28;
+}
```
anuraghazra, line 49: pretty sure we can do this

```ts

function getMaxDay(month: number) {
  return getDaysInMonth(month);
}
```

src/utils/date.ts
```diff
@@ -0,0 +1,74 @@
...
+  const date = dateValue.split("-");
+  if (+date[0] <= 0 || +date[1] < 1 || +date[1] > 12) return;
+
+  const maxDay = isLeapYear(+date[0])
+    ? getMaxDay(+date[1], true)
+    : getMaxDay(+date[1]);
+  if (+date[2] < 1 && +date[2] > maxDay) return;
+
+  return new Date(
+    +date[0],
```
anuraghazra, line 20: We can probably use

```ts
export function parseDate(dateValue: string | undefined) {
  if (!dateValue) return;
  return parseISO(dateValue as string);
}
```

src/calendar/CalendarCellButton.ts
navin-moorthy: let's fix to `isInvalidDateRange`

anuraghazra: Yeah ok but we also have something called "DateRangeValue" type which could be confusing with it...

And this function is making sure that the date is in range.. so maybe "isDateInBetween" or "DateIsInRange" something like that

src/utils/date.ts
navin-moorthy: This did not strike me, I was actually doing this in the stories 🤦

navin-moorthy: Also did the same to stringifyTime()

```js
export function stringifyTime(date: Date) {
  return format(date, "hh:mm:ss");
}
```

src/utils/date.ts
navin-moorthy: I will test them & will let you know the result !!

navin-moorthy: Seems `getDaysInMonth` requires a date but we dont have one.

We only had year & month, hence I used their script - https://github.com/date-fns/date-fns/blob/v2.0.0/src/getDaysInMonth/index.js

```js
function getMaxDay(year: number, month: number) {
  const lastDayOfMonth = new Date(0);
  lastDayOfMonth.setFullYear(year, month, 0);
  lastDayOfMonth.setHours(0, 0, 0, 0);
  return lastDayOfMonth.getDate();
}
```

src/utils/date.ts
navin-moorthy: `parseISO` doesn't accept a value like this 

```html
<input type="date" id="start" name="trip-start" value="201800-07-22">
```

Which is valid according to https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#dates

image

So I used this instead,

```js
export function parseDate(dateValue: string | undefined) {
  if (dateValue == null) return;

  const parsedDate = parse(dateValue, "yyyyyy-MM-dd", new Date());

  // Check for Invalid Date
  if (isNaN(+parsedDate)) return;
  return parsedDate;
}
```

anuraghazra: Ok nice!

navin-moorthy: We can actually do the same for 

```js
// As per https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#times
export function parseTime(timeValue: string | undefined) {
  if (timeValue == null || !isString(timeValue)) return;

  const timeRegex = timeValue.match(/(\d\d)(:(\d\d)){1,2}/i);
  if (timeRegex == null) return;

  const time = timeValue.split(":");
  const date = new Date();

  if (+time[0] < 0 || +time[0] > 23 || +time[1] < 0 || +time[1] > 59) return;

  date.setHours(+time[0]);
  date.setMinutes(+time[1]);
  date.setSeconds(0);

  if (time[2]) {
    if (+time[2] < 0 || +time[2] > 60) return;
    date.setSeconds(+time[2]);
  }

  return date;
}
```

Since we acceps both HH:mm and HH:mm:ss, we need to chekc the regex & split them and then do the parse & check for invalid date.

Above will be more equivalent to our code in the lines of code. So I believe we no need any refactoring here.

src/calendar/CalendarCellButton.ts
navin-moorthy: IMO they are not confusion since `isInvalidDateRange` it has a keyword `is`.

We can stick it to that.

anuraghazra (approved):

PR #89, opened 2020-10-13
test: added tests for button

anuraghazra:


navin-moorthy:
I thought of removing react-aria component & use them as hooks in UI if necessary.

No need to bloat them in our build 🤔

anuraghazra:
Oh i see. so what about the interactions?

navin-moorthy:
> Oh i see. so what about the interactions?

We can remove them aswell, what do you say?

We are just using their hook & creating a Reakit component out of it.
Since we are now more concerned about the bundle size, we can use them directly in the UI as hook when needed.
That's my thought.

anuraghazra:
> > Oh i see. so what about the interactions?
> 
> We can remove them aswell, what do you say?
> 
> We are just using their hook & creating a Reakit component out of it.
> Since we are now more concerned about the bundle size, we can use them directly in the UI as hook when needed.
> That's my thought.

I'm on favor of this approach from the start.

navin-moorthy:
> > > Oh i see. so what about the interactions?
> > 
> > 
> > We can remove them aswell, what do you say?
> > We are just using their hook & creating a Reakit component out of it.
> > Since we are now more concerned about the bundle size, we can use them directly in the UI as hook when needed.
> > That's my thought.
> 
> I'm on favor of this approach from the start.

I just thought we could leverage this from our library.
But now I understand that we keep things separate as much as possible.

navin-moorthy:
I think you can close this PR and I appreciate your work on tests 👏

anuraghazra:
Closing this since we will be removing buttons and interactions

PR #90, opened 2020-10-13
Improved Accordion

navin-moorthy:
- API is based on Reakit's Tab component(Composite & Disclosure)
- Arrow Navigation from the idea of Segment component
- Fixed issues with compose hooks for better accordion usage
- Separate state for simple accordion & allowMultiple accordion.
- Better types with comments & default
- Updated Storybook stories
- Update Tests for the new API
- Handled most of the edge cases for allowToggle & allowMultiple including thier respective wai-aria requirement


src/accordion/AccordionItem.tsx
```diff
@@ -1,84 +1,150 @@
...
+            //  the current Id in composite hence thats handled directly in state
+            options.select(options.id);
+            return;
+          }
+
+          options.unSelect(options.id);
+          return;
+        }
+
+        options.select?.(options.id);
```
anuraghazra, line 111: This bit of code is same as the OnClick handler's, extract it out.

navin-moorthy: I thought you will say this 😆

anuraghazra: Yeah lol 😆 My eyes & brain screams when they see repetitive code which can be refactored :p

src/accordion/AccordionPanel.ts
```diff
@@ -1,55 +1,140 @@
...
+  }
+
+  const panelIndex = getPanelIndex(options.panels, panel);
+  const accordionsWithoutPanel = getAccordionsWithoutPanel(
+    options.items,
+    options.panels,
+  );
+  return accordionsWithoutPanel[panelIndex]?.id || undefined;
+}
+
```
anuraghazra, line 89: Move these functions to the end since isPanelVisible is also on the end.

navin-moorthy: Okay 👍

anuraghazra (approved):

PR #96, opened 2020-10-16
refactor(storybook): ♻️  easier access to stories - removing sub folders

navin-moorthy:
Easier access to stories

image

### DIFF
--- a/src/progress/stories/CircularProgress.stories.tsx
+++ b/src/progress/stories/CircularProgress.stories.tsx
@@ -9,7 +9,7 @@ import {
 } from "./storybook-progress-utils";
 
 export default {
-  title: "Component/Progress/Circular",
+  title: "Progress Circular",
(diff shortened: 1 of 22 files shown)


src/progress/stories/CircularProgress.stories.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from "./storybook-progress-utils";
 
 export default {
-  title: "Component/Progress/Circular",
+  title: "Progress Circular",
```
anuraghazra, line 12: I think we can keep them colocate. it will be easier to browse them and make the tree smaller overall.

navin-moorthy: Okay Progress/Circular.

anuraghazra (approved):

PR #97, opened 2020-10-16
refactor(reakit): ♻️  remove external deps in 3 components

navin-moorthy:
Only use reakit & local imports in 
- Accordion
- Link
- Breadcrumbs

Added JSDoc for Options & added Snapshot for all three instead of inline snapshots to make the test file clean.


src/accordion/__tests__/Accordion.test.tsx
```diff
@@ -32,12 +32,6 @@ const AccordionComponent = (props: any) => {
 };
 
 describe("Accordion", () => {
-  it("should render correctly", () => {
-    const { asFragment } = render(<AccordionComponent />);
```
anuraghazra: pass the baseId={"myid"} so it does not change everytime the component is rendered

navin-moorthy: We need to pass the base id in the state & for each panels.

anuraghazra (approved):

PR #99, opened 2020-10-16
Meter others refactor

navin-moorthy:
Improved Meter & stories


src/accordion/stories/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
...
+import "./index.css";
+import {
+  useAccordionState,
+  Accordion,
+  AccordionTrigger,
+  AccordionPanel,
+  AccordionState,
+} from "../index";
+
+export type AccordionComponentInitialState = {
```
anuraghazra, line 12: We could use "AccordionInitialState" from the AccordionState.ts right?

src/accordion/stories/Accordion.tsx
```diff
@@ -0,0 +1,200 @@
+import * as React from "react";
```
anuraghazra, line 1: Can we rename this file to AccordionComponent.ts ?? since it's confusing because we have the Accordion.ts too

navin-moorthy: I have done the same to Meter & other files I have worked on.
I will change them in the next PR which I forked from this branch

src/accordion/stories/Accordion.tsx
navin-moorthy: Agreed, we can use it.

src/accordion/stories/Accordion.tsx
navin-moorthy: I have already made changes to the same file, changing now and merging will give extra job.

anuraghazra (approved):

PR #103, opened 2020-10-19
Progress improvement

navin-moorthy:


src/meter/Meter.ts
```diff
@@ -24,8 +16,9 @@ const useMeter = createHook<MeterOptions, MeterHTMLProps>({
   compose: useBox,
   keys: METER_KEYS,
 
-  useProps(options, { "aria-valuetext": ariaValueText, ...htmlProps }) {
-    const { value, max, min, percent } = options;
+  useProps(options, htmlProps) {
+    const { value, max, min, ariaValueText } = options;
+    console.log("%c ariaValueText", "color: #99adcc", ariaValueText);
```
anuraghazra, line 21: remove these console logs please.

navin-moorthy: I am about to push a change removing this & more

src/progress/ProgressState.ts
```diff
@@ -4,31 +4,83 @@
...
+export type ProgressInitialState = Partial<
+  Pick<
+    ProgressState,
+    "value" | "min" | "max" | "isIndeterminate" | "ariaValueText"
+  >
+>;
 
-  const percent = value != null ? valueToPercent(value, min, max) : undefined;
-  const isIndeterminate = isUndefined(value);
+export type ProgressStateReturn = ProgressState & ProgressAction;
```
anuraghazra, line 60: `export type ProgressStateReturn = ReturnType<typeof useProgressState>;` would be nicer i guess

anuraghazra: Well yeah we would not get the jsdocs in return state maybe we can keep it this way... i'll do some research to see if tsdoc supports any shortcuts for this with ReturnType generic

navin-moorthy: I am thinking this is nice and better, both does the same.
Helps with clear separation between the state & Actions

anuraghazra: Too much manual types management. ReturnType is easier to manage since it would update depending on the hook's return values.

src/progress/stories/LinearProgress.stories.tsx
```diff
@@ -1,89 +1,31 @@
...
-        {...progress}
-        style={{ ...progressBarStyle, backgroundColor: "none" }}
-        className={indeterminateStyles}
-      />
-    </div>
-  );
+export const WithStripeAnimation = Base.bind({});
+WithStripeAnimation.args = {
+  withStripe: true,
+  withStripeAnimation: true,
```
anuraghazra, line 27: we can inline it 🤔 

```js

export const Default = Base.bind({});

export const WithLabel = Base.bind({}, { withLabel: true });

export const WithStripe = Base.bind({}, { withStripe: true });

export const WithStripeAnimation = Base.bind(
  {},
  {
    withStripe: true,
    withStripeAnimation: true,
  },
);

export const IsIndeterminate = Base.bind({}, { isIndeterminate: true });
```

navin-moorthy: Controls will not work if we do that :(

anuraghazra: Oh ok okay, then no changes needed

src/progress/ProgressState.ts
navin-moorthy: I am just prefering it the Reakit's way. Yeah, Typescript will infer from the returns.
Since we already have the State & Actions defined, I prefered this useful.

This was helpful, when I am improving a featurn where I return more items, Typescript warns me to include those state/Action, so that I can respective JSDoc.

anuraghazra: Okay agreed we can keep it. Just that did not liked the idea of deriving both the `initial props` and `return state` from the same types, for this small example with progress it is fine, but it would be a little bit messy if we wanted to change/rename return values suppose we want to change the "setValue" to "setProgressValue" then we have to overwrite the Return types anyways.


It is fine for this progress component but not very effective for other components like Segment, or Calendar. where props and state are very different.

navin-moorthy: Then we can define the `setProgressValue` in Action and it will be in the Return type.
Even I find `export type ProgressStateReturn = ReturnType<typeof useProgressState>;` this was better initially.

But on deep diving into the motivation of Reakit's approach, I find the latter approach useful. Even for ComboBox reakit has done the same.

anuraghazra: Ok we can keep it no change required

anuraghazra (approved):

PR #104, opened 2020-10-20
chore(jsdoc): toast jsdoc and improvements

anuraghazra:


src/utils/index.ts
```diff
@@ -45,10 +45,9 @@ export function isTouch() {
   );
 }
 
-export const dataAttr = (condition: boolean | undefined) =>
+export const dataAttr = (condition?: boolean) =>
   (condition ? "" : undefined) as Booleanish;
 
-export const ariaAttr = (condition: boolean | undefined) =>
-  condition ? true : undefined;
+export const ariaAttr = (condition?: boolean) => (condition ? true : undefined);
```
navin-moorthy, line 51: Why this change 🤔 , any reasons?

I think line 51 can remain the same.

anuraghazra: Is this intentional that you used type `boolean | undefined`??

because the difference between `boolean | undefined` vs `condtion?: boolean` is first one expects to be undefined or boolean value and the argument cannot be empty.

but `condition?: boolean` argument can be empty and can pass undefined.

image
image

navin-moorthy: Based on the usage, both should be `boolean | undefined`.

We need the `condition` in both utils to be passed

navin-moorthy: I made a mistake, just now gave it a thought !!

anuraghazra: boolean | undefined is totally fine if you intentionally want to restrict the arguments but both would not effect any logic behaviour since if we don't provide the argument it will be undefined anyways.

navin-moorthy: Yeah, it would not affect the behaviour, we are just making it compulsory to pass a condition.

src/toast/ToastProvider.tsx
```diff
@@ -3,6 +3,7 @@ import ReactDOM from "react-dom";
 import { canUseDOM } from "reakit-utils";
 import { createContext } from "@chakra-ui/utils";
```
navin-moorthy, line 4: Extract the utils from chakra - https://github.com/chakra-ui/chakra-ui/blob/develop/packages/utils/src/react-helpers.ts 🤔 to our local utils?

src/toast/ToastController.tsx
```diff
@@ -2,10 +2,29 @@ import React from "react";
 import { useTimeout } from "@chakra-ui/hooks";
```
navin-moorthy, line 2: Can we use their hooks package OR should we also localize this ?

navin-moorthy: Can we reexport form chakra/hook from our newly created "hooks" folder 🤔

anuraghazra: Depends on if we are using chakra hooks anywhere else, if this is the only time we are using the package we can extract it out

navin-moorthy: We are using many hooks from chakra.

anuraghazra: Then i think we can keep it, since it would be well tested & updated.

navin-moorthy: Okay, okay !!

src/toast/ToastProvider.tsx
navin-moorthy: What about this @anuraghazra  ?

src/toast/ToastController.tsx
anuraghazra: Yeah

src/toast/ToastProvider.tsx
anuraghazra: I will extract it out...

navin-moorthy (approved):
LGTM 🎉

PR #107, opened 2020-10-21
Progress & Meter State changes

navin-moorthy:
Update Progress & Meter State too based on Aria spec


navin-moorthy:
There is going to be multiple clamp/clampValues.

I will unify everything once slider is done based on the usage.

navin-moorthy:
Conflicts can be fixed after the approval, because master can be udpated before the review gets completed.

anuraghazra (approved):

PR #109, opened 2020-10-22
test: added tests for live announcer

anuraghazra:
Never got an use case to understand and see how `useImperativeHandle` works, but today I learned about it! 😄


navin-moorthy (approved):
LGTM 🎉

PR #112, opened 2020-10-22
Multi Slider

navin-moorthy:


src/hooks/useItems/useSortBasedOnDOMPosition.ts
```diff
@@ -0,0 +1,60 @@
...
+
+function setItemsBasedOnDOMPosition(items: Item[], setItems: SetItems) {
+  const sortedItems = sortBasedOnDOMPosition(items);
+  if (items !== sortedItems) {
+    setItems(sortedItems);
+  }
+}
+
+// istanbul ignore next: JSDOM doesn't support IntersectionObverser
+// See https://github.com/jsdom/jsdom/issues/2032
```
anuraghazra, line 17: **Note to me when testing this component**

src/slider-new/SliderState.ts
```diff
@@ -0,0 +1,323 @@
...
+
+function replaceIndex<T>(array: T[], index: number, value: T) {
+  if (array[index] === value) {
+    return array;
+  }
+
+  return [...array.slice(0, index), value, ...array.slice(index + 1)];
+}
+
+function clampValues(values: number[] | undefined, min: number, max: number) {
```
anuraghazra, line 316: omg, another one. this one is bit different i think we should name it something else, "bulkClamp" or something like that

src/slider-new/SliderState.ts
```diff
@@ -0,0 +1,323 @@
...
+  if (!values) return [getOptimumValue(min, max)];
```
anuraghazra, line 317: Also it seems that we are returning Optimum value if no values is provided, but i do not feel like this logic belongs here in this function. 

I don't care much about the `Single Responsibility Principle` but IMO this logic should be outside.

src/slider-new/SliderThumb.ts
```diff
@@ -0,0 +1,153 @@
...
+  keys: SLIDER_THUMB_KEYS,
+
+  useProps(options, htmlProps) {
+    const { index, inputs } = options;
+    const isVertical = options.orientation === "vertical";
+    const { direction } = useLocale();
+    const { addGlobalListener, removeGlobalListener } = useGlobalListeners();
+
+    // Immediately register editability with the state
+    options.setThumbEditable(index, !options.isDisabled);
```
anuraghazra, line 50: I think we should put this inside an useEffect to ensure it runs only once when the component mounts, not on each render.

```js
    React.useEffect(() => {
      // Immediately register editability with the state
      options.setThumbEditable(index, !options.isDisabled);

       // maybe put options.isDisabled in the deps
      // eslint-disable-next-line react-hooks/exhaustive-deps
    }, []);
```

src/slider-new/helpers/useMove.ts
```diff
@@ -0,0 +1,246 @@
+import { disableTextSelection, restoreTextSelection } from "./textSelection";
+import React, { HTMLAttributes, useMemo, useRef } from "react";
+import { useGlobalListeners } from "./useGlobalListeners";
+import { isNull } from "../../utils";
+
+export interface BaseMoveEvent {
+  pointerType: "mouse" | "pen" | "touch" | "keyboard";
+}
```
anuraghazra, line 8: We can clean it up a bit.

```ts

export type PointerTypes = "mouse" | "pen" | "touch" | "keyboard";

export interface BaseMoveEvent {
  pointerType: PointerTypes;
}
```

```diff
+ const end = (pointerType: PointerTypes) => {
- const end = (pointerType: BaseMoveEvent["pointerType"]) => {

```

src/slider-new/helpers/useMove.ts
```diff
@@ -0,0 +1,246 @@
...
+
+export function useMove(props: MoveProps): HTMLAttributes<HTMLElement> {
+  const { onMoveStart, onMove, onMoveEnd } = props;
+
+  const state = useRef<{
+    didMove: boolean;
+    lastPosition: { pageX: number; pageY: number } | null;
+    id: number | null;
+  }>({ didMove: false, lastPosition: null, id: null });
```
anuraghazra, line 39: Extract this type out, just looks a bit cleaner.

```ts

interface useMoveStateRef {
  didMove: boolean;
  lastPosition: { pageX: number; pageY: number } | null;
  id: number | null;
}


  const state = useRef<useMoveStateRef>({
    didMove: false,
    lastPosition: null,
    id: null,
  });
```

src/slider-new/helpers/useMove.ts
```diff
@@ -0,0 +1,246 @@
...
+          e.stopPropagation();
+          triggetKeyboardMove(0, -1);
+          break;
+        case "Down":
+        case "ArrowDown":
+          e.preventDefault();
+          e.stopPropagation();
+          triggetKeyboardMove(0, 1);
+          break;
+      }
```
anuraghazra, line 232: Can we use a fallthrough switch case?

```js
      switch (e.key) {
        case "Up":
        case "Left":
        case "Down":
        case "Right":
        case "ArrowUp":
        case "ArrowLeft":
        case "ArrowDown":
        case "ArrowRight":
          e.preventDefault();
          e.stopPropagation();
        case "Left":
        case "ArrowLeft":
          triggetKeyboardMove(-1, 0);
          break;
        case "Right":
        case "ArrowRight":
          triggetKeyboardMove(1, 0);
          break;
        case "Up":
        case "ArrowUp":
          triggetKeyboardMove(0, -1);
          break;
        case "Down":
        case "ArrowDown":
          triggetKeyboardMove(0, 1);
          break;
      }
```

src/slider-new/helpers/useMove.ts
```diff
@@ -0,0 +1,246 @@
...
+            e.pageY - state.current.lastPosition.pageY,
+          );
+          state.current.lastPosition = { pageX: e.pageX, pageY: e.pageY };
+        }
+      };
+
+      const onPointerUp = (e: PointerEvent) => {
+        if (e.pointerId === state.current.id) {
+          // @ts-ignore
+          const pointerType: BaseMoveEvent["pointerType"] =
```
anuraghazra, line 176: const pointerType = (e.pointerType || "mouse") as PointerTypes;

src/slider-new/helpers/useMove.ts
```diff
@@ -0,0 +1,246 @@
...
+          addGlobalListener(window, "mousemove", onMouseMove, false);
+          addGlobalListener(window, "mouseup", onMouseUp, false);
+        }
+      };
+
+      const onTouchMove = (e: TouchEvent) => {
+        // @ts-ignore
+        const touch = [...e.changedTouches].findIndex(
+          ({ identifier }) => identifier === state.current.id,
+        );
```
anuraghazra, line 113: This should also work.

```ts
        const touch = Array.from(e.changedTouches).findIndex(
          ({ identifier }) => identifier === state.current.id,
        );
```

src/slider-new/helpers/useMove.ts
```diff
@@ -0,0 +1,246 @@
...
+          const { pageX, pageY } = e.changedTouches[touch];
+          if (isNull(state.current.lastPosition)) return;
+          move(
+            "touch",
+            pageX - state.current.lastPosition.pageX,
+            pageY - state.current.lastPosition.pageY,
+          );
+          state.current.lastPosition = { pageX, pageY };
+        }
+      };
```
anuraghazra, line 124: I have no idea if this code would work, i'm not able to correctly test it since the touch behaviour is broken in my side but we can try it later.

```ts
      const onTouchMove = (e: TouchEvent) => {
        // const touch = Array.from(e.changedTouches).findIndex(
        //   ({ identifier }) => identifier === state.current.id,
        // );
        const touch = e.changedTouches.item(state.current.id as number);
        if (touch) {
          const { pageX, pageY } = touch;
          if (isNull(state.current.lastPosition)) return;
          move(
            "touch",
            pageX - state.current.lastPosition.pageX,
            pageY - state.current.lastPosition.pageY,
          );
          state.current.lastPosition = { pageX, pageY };
        }
      };
```

src/slider-new/SliderState.ts
navin-moorthy: Yeah it should be as we discussed. I will rename it and refactor it.

src/slider-new/SliderThumb.ts
navin-moorthy: Yeah, we can do that 🤔 
I will check and apply the refactor.

src/slider-new/helpers/useMove.ts
navin-moorthy: We can use this, looks nice !!

src/slider-new/helpers/useMove.ts
navin-moorthy: I believe these won't be necessary.
Reakit will expose useMove on the next release.
We can import them directly, untill then I think it can stay as it is 🤔

src/slider-new/helpers/useMove.ts
navin-moorthy: I think we don't have to make these changes in our repo.
You can raise a PR directly to React Aria.

src/accordion/AccordionState.ts
```diff
@@ -138,7 +137,7 @@ export function useAccordionState(
     [composite.move],
   );
 
-  const panels = useItems();
+  const panels = useCompositeState();
```
anuraghazra, line 140: hmm i think it should be 

```js
  const { items: panel } = useCompositeState();
```

anuraghazra (approved):

PR #115, opened 2020-10-23
test: added tests for createContext

anuraghazra:


src/utils/utils.test.tsx
```diff
@@ -1,11 +1,15 @@
+import React from "react";
+import MockDate from "mockdate";
+import { render } from "reakit-test-utils";
+import { valueToPercent } from "@chakra-ui/utils";
```
navin-moorthy, line 4: import valueToPercent from out utils itself.

navin-moorthy (approved):
LGTM 🎉

src/utils/utils.test.tsx
anuraghazra: Opps.

anuraghazra: VSCode auto imported

PR #129, opened 2020-10-30
chore: wip runtime ts transpile examples

anuraghazra:


.storybook/main.ts
```diff
@@ -2,6 +2,7 @@ module.exports = {
   stories: ["../src/**/*.stories.@(js|jsx|ts|tsx)"],
   addons: [
     "storybook-addon-preview",
+    "@storybook/addon-docs",
```
navin-moorthy, line 5: To suppress warning in Storybook cli logs.
addon-docs must be loaded before controls

scripts/transpile-ts.js
```diff
@@ -0,0 +1,22 @@
+import * as ts from "typescript";
```
navin-moorthy, line 1: This import gives a webpack error, hence added a plugin to suppress that error

PR #136, opened 2020-11-03
Storybook Preview for more components

navin-moorthy:

PR #137, opened 2020-11-04
chore: storybook migrate segment, timepicker, select, toast

anuraghazra:


src/segment/stories/Segment.stories.tsx
```diff
@@ -1,40 +1,49 @@
...
+        codesandbox: DEFAULT_REACT_CODESANDBOX(["renderless-components@alpha"]),
+      },
+      {
+        tab: "React",
+        template: appTemplate,
+        language: "tsx",
+        copy: true,
+        codesandbox: DEFAULT_REACT_CODESANDBOX(["renderless-components@alpha"]),
+      },
+      {
```
anuraghazra, line 28: we will use the createPreviewTabs helper after merging #136

navin-moorthy: 👍

scripts/create-preview-tabs.ts
```diff
@@ -46,3 +47,30 @@ export function createPreviewTabs(props: Props) {
...
+  return {
+      framework: "${ReactTab.toLowerCase()}",
+      files: {
+          "src/App.${type}": previews["${ReactTab}"][0],
+          "src/styles.css": previews["CSS"] ? previews["CSS"][0] : "",
+          "src/ToastUtils.component.${type}": previews["${utilTab}"] ? previews["${utilTab}"][0] : "",
+      },
+      userDependencies: ${joinStrs(deps)},
+  };`) as CodeSandboxTemplate;
+};
```
anuraghazra, line 76: Tried using normal higher order function but does not seem to be working. throwing `type` not defined error. i think internally we loose the function context and i think thats why addon-preview also uses string template literal functions

navin-moorthy: Oh okay okay !!

navin-moorthy (approved):
LGTM 🎉

navin-moorthy:
Changes looks good, we can merge it after the #139 I think including those changes as well.

PR #142, opened 2020-11-06
useControllable on all components

navin-moorthy:


src/accordion/AccordionState.ts
```diff
@@ -5,33 +5,32 @@ import {
   CompositeInitialState,
 } from "reakit";
 import * as React from "react";
-import {
-  SealedInitialState,
-  useSealedState,
-} from "reakit-utils/useSealedState";
+import { useControllableState } from "@chakra-ui/hooks";
 
 export type AccordionState = CompositeState & {
```
anuraghazra, line 10: let's not self reference the types for the accordion actions. just to reduce confusion. 

```diff
- registerPanel: AccordionActions["registerItem"];
+ registerPanel: CompositeActions["registerItem"];
```

```ts

export type AccordionActions = CompositeActions & {
  /**
   * Moves into and selects an accordion by its `id`.
   */
  select: CompositeActions["move"];
  /**
   * Moves into and unSelects an accordion by its `id` if it's already selected.
   */
  unSelect: CompositeActions["move"];
  /**
   * Sets `selectedId`.
   */
  setSelectedId: React.Dispatch<React.SetStateAction<string | null>>;
  /**
   * Sets `selectedIds`.
   */
  setSelectedIds: React.Dispatch<React.SetStateAction<(string | null)[]>>;
  /**
   * Registers a accordion panel.
   */
  registerPanel: CompositeActions["registerItem"];
  /**
   * Unregisters a accordion panel.
   */
  unregisterPanel: CompositeActions["unregisterItem"];
};
``

File: src/number-input/NumberInputState.ts [hunk 2]
```diff
@@ -149,116 +148,37 @@ export type NumberinputInitialState = Pick<
...
-
-  const reset = React.useCallback(() => {
-    let next: StringOrNumber;
-    if (initialValue == null) {
-      next = "";
-    } else {
-      next = castValue(initialValue, stepProp, precisionProp);
-    }
-    setValue(next);
-  }, [initialValue, precisionProp, stepProp, setValue]);
```
anuraghazra, line 253: Satisfying! nice!

anuraghazra (approved):

src/number-input/NumberInputState.ts
navin-moorthy: Why wouldn't you be 😅

PR #143, opened 2020-11-09
feat: accordion types improvements

anuraghazra:


src/accordion/AccordionPanel.ts
```diff
@@ -10,21 +10,17 @@ import * as React from "react";
...
-    | "selectedIds"
-    | "registerPanel"
-    | "unregisterPanel"
-    | "panels"
-    | "items"
-    | "allowMultiple"
-  > & {
+    "registerPanel" | "unregisterPanel" | "panels" | "items" | "allowMultiple"
+  > &
+  SelectedIdPair & {
```
navin-moorthy, line 23: Pick only the selectedId & selectedIds

src/accordion/AccordionState.ts
```diff
@@ -1,117 +1,73 @@
...
-    defaultSelectedId = null,
-    onSelectedIdChange,
-    selectedIds: selectedIdsProp,
-    defaultSelectedIds = [],
-    onSelectedIdsChange,
-    allowMultiple = false,
-    allowToggle: allowToggleProp = false,
     manual = true,
+    allowToggle: allowToggleProp = false,
+    allowMultiple,
```
navin-moorthy, line 43: Not needed !

src/accordion/types.ts
```diff
@@ -0,0 +1,168 @@
...
+  MultiReturn;
+
+type AccordionProps = SingleAccordionProps | MultiAccordionProps;
+export type AccordionReturns = MultiReturn | SingleReturn;
+
+export type AccordionInitialState = CompositeInitialState & AccordionProps;
+export type AccordionInitialStateSingle = CompositeInitialState &
+  SingleAccordionProps;
+export type AccordionInitialStateMulti = CompositeInitialState &
+  MultiAccordionProps;
```
navin-moorthy, line 168: What about setting <Partial> here instead of doing it inside the components 🤔

src/accordion/stories/AccordionBasic.component.tsx
```diff
@@ -5,10 +5,10 @@ import {
   AccordionPanel,
   AccordionTrigger,
   useAccordionState,
-  AccordionInitialState,
 } from "renderless-components";
+import { AccordionInitialState } from "../types";
```
navin-moorthy, line 9: Not Needed !!

anuraghazra: i've currently set any in the App prop... so will add it back since storybook is forcing the allowMultiple to be true

src/accordion/types.ts
anuraghazra: Can you be specific? what setting?

src/accordion/AccordionState.ts
anuraghazra: we need to do this then since it needed to be fallback to false

image

navin-moorthy: I am just saying we can remove line 43, since we are using props.allowMultiiple everywhere.

src/accordion/types.ts
navin-moorthy: Oops, that should be setting `<Partial>`.

I did not put them in code blocks 😅

src/accordion/stories/AccordionBasic.component.tsx
navin-moorthy: Oh, right right !!

src/accordion/types.ts
anuraghazra: Yeah i think we can put in partial here i left it since i was doing experiments and wanted full control

navin-moorthy (approved):
LGTM 🎉

PR #144, opened 2020-11-09
Pickers chore

navin-moorthy:


src/calendar/stories/RangeCalendar.stories.tsx
```diff
@@ -71,24 +71,22 @@ export const Default = Base.bind({});
...
 
 export const Options = Base.bind({});
 Options.args = {
-  start: new Date(),
-  end: addWeeks(new Date(), 1),
+  defaultStart: new Date(),
+  defaultEnd: addWeeks(new Date(), 1),
   minValue: null,
   maxValue: null,
   isDisabled: false,
```
anuraghazra, line 92: If this is intentional then it's fine but using default values means we cannot change the values from the storybook control addon

src/segment/Segment.ts
```diff
@@ -95,99 +94,116 @@ export const useSegment = createHook<SegmentOptions, SegmentHTMLProps>({
...
+            break;
           }
+        }
+      },
+      [enteredKeys, next, options, segment.maxValue, segment.type],
+    );
 
-          options.setSegment(segment.type, segmentValue);
+    const onKeyDown = React.useCallback(
+      (e: any) => {
```
anuraghazra, line 148: (e: React.KeyboardEvent)

anuraghazra (approved):

src/calendar/stories/RangeCalendar.stories.tsx
navin-moorthy: Oh, right right !! I did not know this !!

I will revert back.

PR #147, opened 2020-11-10
Combobox & Select

navin-moorthy:


src/select/helpers/types.ts
```diff
@@ -0,0 +1,12 @@
...
+  ref: React.RefObject<HTMLElement>;
+};
+
+export type Item = {
+  id: string | null;
+  ref: React.RefObject<HTMLElement>;
+  groupId?: Group["id"];
+  disabled?: boolean;
+  value?: string;
+};
```
anuraghazra, line 12: Reakit already has these types so we can probably do this

```ts
import {
  Item as ReakitItem,
  Group as ReakitGroup,
} from "reakit/ts/Composite/__utils/types";

export type Group = ReakitGroup;
export type Item = ReakitItem & {
  value?: string;
};

```

src/select/SelectBaseState.ts
```diff
@@ -0,0 +1,176 @@
...
+    () => getValueFromId(valuesById.current, composite.currentId),
+    [valuesById.current, composite.currentId],
+  );
+
+  const items = React.useMemo(() => {
+    compositeItems.forEach(item => {
+      if (item.id) {
+        (item as Item).value = valuesById.current[item.id];
+      }
+    });
```
anuraghazra, line 64:
```ts
    compositeItems.forEach((item: Item) => {
      if (item.id) {
        item.value = valuesById.current[item.id];
      }
    });
```

src/select/SelectBaseState.ts
```diff
@@ -0,0 +1,176 @@
...
+  valuesById: ValuesById,
+  currentId: CompositeState["currentId"],
+) {
+  if (currentId == null) return;
+
+  const selectedOption = valuesById.find(option => option.id === currentId);
+  return selectedOption ? selectedOption.value : undefined;
+}
+
+export function useSelectBaseState<T extends CompositeStateReturn>(
```
anuraghazra, line 29: Is this taken from reakit codebase? any particular use case for using a generic for this?

src/select/Select.ts
```diff
@@ -1,31 +1,115 @@
...
+            options.values,
+            character,
+            item => item ?? "",
+            selectedValue,
+          );
+
+          if (nextItem) {
+            options.setSelectedValue(nextItem);
+          }
+        }),
```
anuraghazra, line 89: Let's just extract this to a `handleTypeahead` function to make it look cleaner

src/select/Select.ts
```diff
@@ -1,31 +1,115 @@
...
+        const keyMap = {
+          Enter: first,
+          " ": first,
+          ArrowUp: last,
+          ArrowDown: first,
+        };
 
-    return htmlProps;
+        const action = keyMap[event.key as keyof typeof keyMap];
+        if (action) action();
```
anuraghazra, line 50: action?.();

src/select/helpers/getItemId.ts
```diff
@@ -0,0 +1,7 @@
+function kebabCase(string: string) {
+  return string.toLowerCase().replace(/[^a-z0-9]/g, "-");
+}
```
anuraghazra, line 3: Let's move this to `utils` folder

src/select/helpers/useTypeaheadShortcut.ts
```diff
@@ -3,46 +3,44 @@ import { closest } from "reakit-utils";
 import { useShortcut } from "@chakra-ui/hooks";
 import { getNextItemFromSearch } from "@chakra-ui/utils";
 
-import { SelectStateReturn } from "./SelectState";
+import { SelectPopoverOptions } from "../SelectPopover";
+import { getIdFromValue } from "../SelectBaseState";
 
-interface usePortalShortcutProps {
+interface useTypeaheadShortcutProps {
+  options: SelectPopoverOptions;
```
anuraghazra, line 10: Let's pick what we need 

```ts
  options: Pick<
    SelectPopoverOptions,
    "values" | "currentValue" | "move" | "valuesById"
  >;
```

src/select/SelectBaseState.ts
navin-moorthy: It's from Combobox.

I am not sure why it's needed, we can also type them directly right ?

anuraghazra: Yeah we can type them directly no need of generics. I think maybe they had 2d and 1d grid combobox that's why they used generic.

navin-moorthy: Right, I removed the grid since it's not needed in this commit - https://github.com/timelessco/renderless-components/pull/147/commits/0637aea6608634f4ffe1adf446aad4751bff5d99

anuraghazra (approved):

PR #152, opened 2020-11-17
build(reakit): 📌 make reakit as a peer deps

navin-moorthy:


src/select/stories/SelectControlled.component.tsx
```diff
@@ -75,3 +74,75 @@ const SelectComp: React.FC<SelectCompProps> = props => {
 };
 
 export default App;
+
+const fruits = [
```
anuraghazra, line 78: What was the issue previously with having this in separate file?

navin-moorthy: We need to add a separate file to the CSB & change the stories 😛

anuraghazra: hmmm.. considering the chore.. i think it would have been better if we just had 1 single file and just changed the stories to have one extra fruit.js file with this content. 🤔  (what we actually did with ToastUtils.component.tsx)

because this looks repetitive to me.

anuraghazra (approved):

PR #159, opened 2020-11-19
fix(select): 🐛  focus the popover on loop from both ends

### LINKED ISSUE #156: [Select] Do not focus Composite when looping through CompositeItem
Reproduction - https://codesandbox.io/s/select-renderless-components-i8cc1?file=/src/App.js

navin-moorthy:
closes #156

### DIFF
--- a/src/select/SelectListState.ts
+++ b/src/select/SelectListState.ts
@@ -14,15 +14,11 @@ import {
 export function useSelectListState(
   initialState: SealedInitialState<SelectListInitialState> = {},
 ): SelectListStateReturn {
-  const {
-    currentId = null,
-    orientation = "vertical",
-    loop = true,
-    ...sealed
-  } = useSealedState(initialState);
+  const { orientation = "vertical", loop = true, ...sealed } = useSealedState(
+    initialState,
+  );
 
   const composite = useCompositeState({
-    currentId,
     orientation,
     loop,
     ...sealed,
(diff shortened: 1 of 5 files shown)


src/select/__tests__/__snapshots__/Select.test.tsx.snap
```diff
@@ -14,6 +14,7 @@ exports[`Select should render correctly 1`] = `
       Select a fruit
     </button>
     <div
+      aria-activedescendant="select-apple"
```
anuraghazra, line 17: Why apple is already in the activedescendant without providing the "selectedValue"?

anuraghazra (approved):

src/select/SelectListState.ts
```diff
@@ -14,15 +14,11 @@ import {
 export function useSelectListState(
   initialState: SealedInitialState<SelectListInitialState> = {},
 ): SelectListStateReturn {
-  const {
-    currentId = null,
```
navin-moorthy, line 18: Thats because I removed this !

Now by default current id will be on the first tabbable element as programmed in composite.

navin-moorthy: This will only affect the currentValue not the selectedValue.

If the selectedValue is changed, currentId will change on useEffect when visible
image

anuraghazra: Got it!

PR #155, opened 2020-11-17
Updated Docs

navin-moorthy:

PR #160, opened 2020-11-19
docs: added props generation script

anuraghazra:
Should be merged with #157 

And 157 should be merged with #155


navin-moorthy (approved):
LGTM 🎉

PR #174, opened 2020-12-02
docs(component): 📝  fix yaml bug and finish component docs

navin-moorthy:


navin-moorthy:
Just the docs update, we can revert it back later if it's not appropriate.

PR #183, opened 2020-12-07
docs(docgen): 📝  use react-docs-typescript

navin-moorthy:
Used react-docgen-typescript !!

I think our current method is more customizable.
Explore futhur here.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,6 +46,7 @@
     "contributors:add": "all-contributors add",
     "contributors:generate": "all-contributors generate",
     "docs": "node scripts/builds/docs.js",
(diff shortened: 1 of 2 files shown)


navin-moorthy:
Do you need to still look into this @anuraghazra ?

anuraghazra:
No I've explored this, it does not actually solves our main issue tho, and our current approach is already fine and is also being used for generating other composition,examples,csblink etc etc so i think we can close this for now.

navin-moorthy:
Implemented an alternated approach, hence closing this.

PR #187, opened 2020-12-11
fix(accordion): undefined discriminated union typeguard

anuraghazra:
Found the fix while fixing the typescript in our tailwind datepicker


navin-moorthy (approved):
LGTM 🎉

PR #190, opened 2021-01-18
test: fix timepicker tests

anuraghazra:


anuraghazra:
Weird build test was failing on my end.. fixed it but now it's failing on CI. seems like a local datetime setting issue

anuraghazra:
Closing this because seems like it was issue with local date system

PR #375, opened 2022-06-29
refactor(slider): ♻️  remove orientation from verticalSlider

mcnaveen:
This PR removes the orientation args from the SingleVertical slider. Because the same can be achieved using the AllInOne slider.


navin-moorthy (approved):
LGTM 🎉

